PR #66, opened 2019-10-09
Add base typography styles

kgabryje:
https://github.com/apache/airflow-site/issues/22


landing-pages/site/assets/scss/_typography.scss
```diff
@@ -0,0 +1,138 @@
...
+$primary-font: 'Roboto', sans-serif;
+$monospace-font: 'Roboto Mono', monospace;
+
+@mixin header($size) {
+  font-family: $header-font;
+  color: $greyish-brown;
+  font-weight: 500;
+
+  @if $size == "large" {
+    font-size: 72px;
```
mik-laj: Personally, I prefer font size in rem, but px is also ok.

PR #89, opened 2019-10-22
Add blog tags

kgabryje:


landing-pages/site/assets/scss/main-custom.scss
```diff
@@ -18,7 +18,7 @@
  */
 
 @import url('https://fonts.googleapis.com/css?family=Rubik:500&display=swap');
-@import url('https://fonts.googleapis.com/css?family=Roboto:400,500,700&display=swap');
+@import url('https://fonts.googleapis.com/css?family=Roboto:400,400i,500,700&display=swap');
```
mik-laj: ?

kgabryje: I added Roboto italic because we have italic style in typography. Though i realise now that it shouldn't be a part of this PR.

mik-laj: It doesn't have to be in a separate PR. I just didn't know what happened here and I thought it was a typo

kgabryje: 👍

PR #90, opened 2019-10-23
Improve accordion spacings and arrow rotation 

kgabryje:


landing-pages/site/layouts/shortcodes/accordion.html
```diff
@@ -25,7 +25,9 @@
         </div>
         <div class="accordion__summary-content">
             {{ if $icon }}
-                {{ with resources.Get $icon }}{{ ( . | minify).Content | safeHTML }}{{ end }}
+                <div class="accordion__summary-content--icon">
+                    {{ with resources.Get $icon }}{{ ( . | minify).Content | safeHTML }}{{ end }}
```
mik-laj: Something is wrong here
Screenshot 2019-10-26 at 14 35 59

kgabryje: Hmm that's weird, because it works well on other pages. For some reason everything below icon gets wrapped in `<pre><code>` tags. Maybe we could ignore it since `Example` page gets deleted anyways?

mik-laj: I fixed it.

PR #93, opened 2019-10-27
Use system hugo to build

mik-laj:
Hello,

We have installed hugo in docker image, but node installs a copy of it.  As it is better to use the official installation packages, I delete the versions from the node. God only knows what's in the packages from the npm registry.

Thanks


kgabryje:
Ok, we just need to remember to make sure that we're using `extended` version of Hugo - otherwise SCSS won't work

PR #95, opened 2019-10-27
Add search blog posts

mik-laj:

PR #96, opened 2019-10-28
Add integration search

mik-laj:


mik-laj:
@kgabryje Can i ask for review?

kgabryje:
Does it work well for you? I see that searchbar is un-styled. Also, there is an error logged from `handleActiveVideo.js`, which I think I fixed in previous commits by changing line 29 from ` if (!videosList) return` to `if (videosList.length === 0) return`

landing-pages/src/js/integrationList.js
```diff
@@ -0,0 +1,153 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import debounce from "lodash/debounce";
+import shuffle from "lodash/shuffle";
+
+const root = document.querySelector("#integrations");
```
kgabryje: Can you add a check if root is not null? It logs errors on pages other than home.

landing-pages/src/js/integrationList.js
```diff
@@ -0,0 +1,153 @@
...
+      (integration) => integration.name.indexOf(keyword) >= 0
+    );
+    return Promise.resolve(selectedIntegration);
+  };
+
+  const filterVisible = (integrations) => {
+    if (currentPage === 1) {
+      integrations.sort(sortByIndex);
+    } else {
+      integrations.sort(sortByName);
```
kgabryje: Do we want to have different order of items after clicking the "show more" button? Isn't it a bit confusing?

mik-laj: I don't know the solution, but it's actually confusing, but posting a random list will also always be misleading, and it's necessary not to promote one integration.

landing-pages/src/js/integrationList.js
mik-laj: Agree. I will add.

mik-laj:
Search  styles are available in other PRs. This PR does not deal with his styles, but with the mechanism.
https://github.com/apache/airflow-site/pull/95
 Other bugs not related to this PR also can't be repaired in this PR, but another one is needed to fix it.

landing-pages/src/js/integrationList.js
kgabryje: I agree, but we achieve randomness by shuffling integrations on line 79. What I mean here is maybe we shouldn't change the type of sorting after clicking button (so it would just be sorted by index).

kgabryje:
I see, thank you for explanation

landing-pages/src/js/integrationList.js
mik-laj: Viewing a long random list will be more confusing than reordering a few items after clicking, but I can change that and ask for feedback tomorrow.

PR #160, opened 2019-11-14
Feature/shuffling commiters

kgabryje:


mik-laj:
@kgabryje Can you do a rebase?

PR #164, opened 2019-11-14
[depends on #151] Feature/share social media

kgabryje:


mik-laj:
One icons is smaller. Can you fix it?
Screenshot 2019-11-15 at 12 58 51

PR #179, opened 2019-11-18
Replace favicons with transparent versions

kgabryje:


ashb:
landing-pages/site/static/favicons/ms-icon-310x310.png  looks blury FYI

kgabryje:
> landing-pages/site/static/favicons/ms-icon-310x310.png looks blury FYI

Updated, thanks for pointing that out!

PR #205, opened 2019-11-28
Fix links to integrations

kgabryje:
Some integrations listed on the home page don't have a corresponding entry in integrations section of docs. @mik-laj @ashb  Could you please verify where the links should point?

### DIFF
--- a/landing-pages/site/static/integrations.json
+++ b/landing-pages/site/static/integrations.json
@@ -1,406 +1,406 @@
 [
   {
     "name": "Apache Cassandra",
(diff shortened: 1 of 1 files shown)


kgabryje:
Now I see that `operators-and-hooks-ref` section to which we link the integration boxes is available only in the latest version of documentation, which we don't provide on our website. In that case should we provide links to readthedocs? Or  do we keep it as it is in this commit https://github.com/apache/airflow-site/pull/205/commits/2589e088b9c69bcadc8bde46085d2a3169e01517 ?

mik-laj:
We should release new documentation with valid document ASAP.

PR #214, opened 2019-11-29
Add debounce to squares positions recalculating

kgabryje:
https://github.com/apache/airflow-site/issues/213


mik-laj:
IT's a quick fix for Safari on iOS. Can you create a issues about full fix?

mik-laj:
I found: https://github.com/apache/airflow-site/issues/213

Issue #122, opened 2019-11-05
Add logo support in the integration list


mik-laj:

kgabryje:
@mik-laj Done?

kaxil:
Done by https://github.com/apache/airflow-site/pull/245

PR #245, opened 2020-02-08
Add logos to integrations list

louisguitton:
_Disclaimer: I am not paid by AWS; I just happened to have the assets at hand; I was curious about how it would look after reading the issue #122; and I use Airflow mostly with AWS at work_ 

In this PR, I add logos for all AWS services.

- I used the assets from https://aws.amazon.com/architecture/icons/
- logos are `.png` and not `svg`
- I use everytime the "Light background" logo as it is the smallest (eg for cloudwatch: light background is 7 KB; regular background is 20 KB)

Although this works and gives a very strong pop to the landing page, it introduces some CSS issues that I will look into and see if I can fix it quickly; otherwise I'd love collaboration on this PR

Before CSS fix: 

Screenshot 2020-02-08 at 13 54 01

image

After CSS fix:
image

cc: @mik-laj @kgabryje


mik-laj:
Can you also update the logos for another provider? I do not want only one supplier to be promoted. Integrations that contain logos are at the top of the list, so they would be more visible than the others.
Here is icons for GCP: https://cloud.google.com/icons

louisguitton:
> Can you also update the logos for another provider?

I added 
- GCP logos using your link
- Apache logos using http://apache.org/logos/#
- Azure logos using https://www.microsoft.com/en-us/download/details.aspx?id=41937

mik-laj:
Now I will call the UI / UX team to check that everything is correct. I am very happy that you did it.

Preview link: https://legal-cough.surge.sh/

kaxil:
Awesome work @louisguitton 🎉 

@kgabryje Any comments?

mik-laj:
@kaxil This person is no longer working on Airflow, but I have a response from the design team. 

> cool, only those gray squares would reduce a bit (but that they would still be centered), because now they are a bit close to the text

It's "cool", so we can merge it.

PR #266, opened 2020-04-17
Improve documentation on DockerOperator

dedunumax:
https://github.com/apache/airflow/issues/8418


zhongjiajie:
Hi, @dedunumax the docs-archive was generate from https://github.com/apache/airflow/tree/master/docs, so I think we should not directly change it, we should change source in https://github.com/apache/airflow/tree/master/docs. Your change will be release in feature, maybe next release.

dedunumax:
Thank you @zhongjiajie! Let me create PR on airflow repository

zhongjiajie:
You are welcome

PR #287, opened 2020-08-25
Docs for 1.10.12

kaxil:


mik-laj (approved):

mik-laj:
https://github.com/apache/airflow-site/blob/docs-for-1.10.12/docs-archive/stable.txt
we have 1..10.12 here so LLGTM.

PR #292, opened 2020-09-17
add geniumcloud in ecosystem

sumyan:
Please review an additions to the list of tools based on Airflow.

### DIFF
--- a/landing-pages/site/content/en/ecosystem/_index.md
+++ b/landing-pages/site/content/en/ecosystem/_index.md
@@ -43,6 +43,8 @@ If you would you like to be included on this page, please reach out to the [Apac
 
 ## Tools integrating with Airflow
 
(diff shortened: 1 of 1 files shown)


turbaszek (approved):

XD-DENG:
This product didn't mention Airflow (implicitly or explicitly) at all in the site. I'm not sure if it's valid to be added here.

XD-DENG (dismissed):
As mentioned in the comment above

sumyan:
Updated the page, please review.

sumyan:
> Updated the page, please review.

sumyan:
> This product didn't mention Airflow (implicitly or explicitly) at all in the site. I'm not sure if it's valid to be added here.

I assure you the product has direct relation to the Airflow, since it's built on top of it. There is no point of adding to the list if it is not related. Please re-check the page again.

PR #293, opened 2020-09-18
Remove duplicate text under Principles section

vin-88:
Following line is written twice under "Principles ->Elegant" section on the Homepage: "Airflow pipelines are lean and explicit".
This change is for fixing the same.


XD-DENG (approved):
LGTM. Thanks!

PR #328, opened 2020-12-03
Automatically remove deleted files from the deploy branch

mik-laj:
Hello,

I had to change the action, which is responsible for publishing the changes on a separate branch, because it did not delete the deleted files. See:  https://github.com/apache/airflow-site/tree/595a17f/docs

Additionally, I used commit hash instead of tags to improve security. Now the author of the action cannot modify the code that will be run.

Thanks to the icon, finding the step is faster because we don't have to read the descriptions.

Success build on master: https://github.com/PolideaInternal/airflow-site/runs/1494326366

Best regards,
Kamil Breguła


potiuk (approved):
I love the ICONS . let's add them to Airlfow :)

kaxil (commented):
Nice

PR #329, opened 2020-12-03
Use Github Acton for pre-commit

mik-laj:
Hello,

This action automatically configures the cache.

Have a purring day!
Kamil Breguła


potiuk (approved):
Nice. Maybe we should add it to airflow too!

PR #331, opened 2020-12-03
Improvements for Github Action Artifacts

mik-laj:
Hello,

I shortened the retention period for artifacts because we don't need them that long. I fixed uploading the theme demo because there was a path and to prevent regression I setup all actions to treat the missing files as an error.

Best regards,
Kamil Breguła


mik-laj:
@potiuk Do you have any idea why Github Action didn't start?

turbaszek (approved):

potiuk (approved):

PR #337, opened 2020-12-09
Update package.json to work with yarn or npm

ashb:
I don't have npm installed so this updates the commands in package.json
to work automatically with whichever of yarn or npm invoked the script.


mik-laj (approved):
I recommend to use ./site.sh, which builds a container that takes care of the installation of all necessary dependencies.

ashb:
> I recommend to use ./site.sh, which builds a container that takes care of the installation of all necessary dependencies.

Yeah, I was debugging a problem/odd behaviour (turned out to be some local changes/old local state/old version) and figured this was a good change to keep anyway.

PR #339, opened 2020-12-09
Add theme options to theme.conf

ashb:
Rather than using custom config properties, and then putting them in to
the theme context, this uses Sphinx's built in mechanism that
automatically exposes every option in `theme.conf` with a `theme_`
prefix.

This commit will make a bit more sense in a second as I am going to add a few more options in here.

I will follow up with a PR to update apache-airflow too.

### DIFF
--- a/sphinx_airflow_theme/sphinx_airflow_theme/theme.conf
+++ b/sphinx_airflow_theme/sphinx_airflow_theme/theme.conf
@@ -6,3 +6,6 @@ pygments_style = default
 [options]
 canonical_url =
 analytics_id =
+# Default set by python code, need to list this here to avoid warning from Sphinx
+navbar_links =
+hide_website_buttons = false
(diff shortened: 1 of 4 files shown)


sphinx_airflow_theme/sphinx_airflow_theme/theme.conf
```diff
@@ -6,3 +6,6 @@ pygments_style = default
 [options]
 canonical_url =
 analytics_id =
+# Default set by python code, need to list this here to avoid warning from Sphinx
+navbar_links =
```
mik-laj, line 10: How can I set a dictionary here? Can you update docs? https://github.com/apache/airflow-site/tree/master/sphinx_airflow_theme#configuration

ashb: Will update docs, you can't specify dict in here (but shpinx warns if this isn't here), so you have to set it via conf.py in end user (airflow)

ashb: Done, PTAL.

mik-laj (approved):

PR #340, opened 2020-12-09
Show hidden ToCs in the sidebar.

ashb:
Relates to apache/airflow#12554

This config, naming, and default behaviour was taken from the
"alabaster" sphinx theme (the default HTML theme)

I have tested that with these default settings, we can put back the `:hidden:` and the sidebar is still populated.

### DIFF
--- a/sphinx_airflow_theme/README.md
+++ b/sphinx_airflow_theme/README.md
@@ -84,6 +84,11 @@ html_theme_options = {
 }
 ```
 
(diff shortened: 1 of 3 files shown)


ashb:
@mik-laj This was the change that I was thinking we needed to make 0.0.2 for.

kaxil (approved):

potiuk (approved):

PR #349, opened 2020-12-11
Sync version between release and package

mik-laj:
When we do version bump, we had to change the release version and package version in setup.py. Now I have synchronized these two values. Just change the value in one place.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -104,7 +104,7 @@ jobs:
           GITHUB_COMMIT: ${{ github.sha }}
           GITHUB_REF: ${{ github.ref }}
         if: >
-          github.ref == 'refs/heads/master' &&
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -108,11 +108,13 @@ jobs:
           github.event_name == 'push'
         run: |
           gh auth status
-          gh release delete v0.0.3 -y || true
+          VERSION="$(cat setup.py | grep 'version=' | cut -d "'" -f 2)"
```
ashb, line 111: If we are okay with "running" setup.py we can do `python setup.py --version`

ashb (approved):

potiuk (approved):

.github/workflows/build.yml
mik-laj: I applied your suggestion.

PR #352, opened 2020-12-12
Fix spelling and remove trailing whitespace

jbampton:
- editor auto strips trailing whitespace


ashb:
These files are autogenerated from sources. Do not edit them.

PR #362, opened 2020-12-21
Fix redirect to / and /docs endpoint for some versions

feluelle:
Some versions like >=1.10.14 and >=2.0.0 don't get correctly redirected to its documentation page.

### DIFF
--- a/landing-pages/site/static/.htaccess
+++ b/landing-pages/site/static/.htaccess
@@ -1,11 +1,9 @@
-RedirectMatch Permanent ^/((1\.10\.[0-9]+|2\.[0-9.]+|stable)/.*)$ "https://airflow.apache.org/docs/apache-airflow/$1"
-RedirectMatch Permanent ^/docs/((1\.10\.[0-9]+|2\.[0-9.]+|stable)/.*)$ "https://airflow.apache.org/docs/apache-airflow/$1"
-
-# Airflow 1.10.x
 RedirectMatch Permanent ^/docs/(stable|1.10.10)/api(\.html)?$ "https://airflow.apache.org/docs/apache-airflow/$1/rest-api-ref"
 RedirectMatch Permanent ^/docs/(stable|1.10.10)/(cli|macros)(\.html)?$ "https://airflow.apache.org/docs/apache-airflow/$1/$2-ref"
 RedirectMatch Permanent ^/((_api|_images|_modules|_sources|_static|howto)/.*)$ "https://airflow.apache.org/docs/apache-airflow/stable/$1"
+RedirectMatch Permanent ^/docs/((1\.10\.[0-9]+|2\.[0-9.]+|stable)/.*)$ "https://airflow.apache.org/docs/apache-airflow/$1"
 RedirectMatch Permanent ^/((api|changelog|cli|concepts|errors|faq|genindex|http-routingtable|installation|integration|kubernetes|license|lineage|macros|metrics|plugins|privacy_notice|profiling|project|py-modindex|scheduler|search|security|start|timezone|tutorial|ui)(\.html)?)$ "https://airflow.apache.org/docs/apache-airflow/stable/$1"
+RedirectMatch Permanent ^/((1\.10\.[0-9]+|2\.[0-9.]+|stable)/.*)$ "https://airflow.apache.org/docs/apache-airflow/$1"
 
 Redirect Permanent /objects.inv https://airflow.apache.org/docs/apache-airflow/stable/objects.inv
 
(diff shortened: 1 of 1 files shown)


feluelle:
I think that should do it.

landing-pages/site/static/.htaccess
```diff
@@ -1,9 +1,11 @@
+RedirectMatch Permanent ^/((1\.10\.[0-9]+|2\.[0-9.]+|stable)/.*)$ "https://airflow.apache.org/docs/apache-airflow/$1"
+RedirectMatch Permanent ^/docs/((1\.10\.[0-9]+|2\.[0-9.]+|stable)/.*)$ "https://airflow.apache.org/docs/apache-airflow/$1"
+
+# Airflow 1.10.x
 RedirectMatch Permanent ^/docs/(stable|1.10.10)/api(\.html)?$ "https://airflow.apache.org/docs/apache-airflow/$1/rest-api-ref"
 RedirectMatch Permanent ^/docs/(stable|1.10.10)/(cli|macros)(\.html)?$ "https://airflow.apache.org/docs/apache-airflow/$1/$2-ref"
```
ashb, line 6: These won't match anymore, as lines 1 and 2 take over. So I think these need to go first.

feluelle: okay, I reverted the order.

landing-pages/site/static/.htaccess
```diff
@@ -1,9 +1,9 @@
...
 RedirectMatch Permanent ^/((_api|_images|_modules|_sources|_static|howto)/.*)$ "https://airflow.apache.org/docs/apache-airflow/stable/$1"
-RedirectMatch Permanent ^/((1.10.1|1.10.2|1.10.3|1.10.4|1.10.5|1.10.6|1.10.7|1.10.8|1.10.9|1.10.10)/.*)$ "https://airflow.apache.org/docs/apache-airflow/$1"
```
feluelle, line 4: So @ashb that > 1.10.10 was missing here was wrong, right?

mik-laj: I think only versions 1.10.1 to 1.10.6.
https://github.com/apache/airflow-site/tree/f18ad1c764861e8b049391c7bf95f46a4adae916%5E

feluelle: I reverted this change as I only want to address the actual issue I found. I think this works as expected.

Only 1.10.14 and >= 2.0.0 was missing in the redirect from / to /docs/apache-airflow.

feluelle: WDYT? @mik-laj - pls re-review :)

turbaszek:
Does this mean we can close this one?
https://github.com/apache/airflow/pull/13250

landing-pages/site/static/.htaccess
```diff
@@ -3,7 +3,7 @@ RedirectMatch Permanent ^/docs/(stable|1.10.10)/(cli|macros)(\.html)?$ "https://
 RedirectMatch Permanent ^/((_api|_images|_modules|_sources|_static|howto)/.*)$ "https://airflow.apache.org/docs/apache-airflow/stable/$1"
 RedirectMatch Permanent ^/((1.10.1|1.10.2|1.10.3|1.10.4|1.10.5|1.10.6|1.10.7|1.10.8|1.10.9|1.10.10)/.*)$ "https://airflow.apache.org/docs/apache-airflow/$1"
 RedirectMatch Permanent ^/((api|changelog|cli|concepts|errors|faq|genindex|http-routingtable|installation|integration|kubernetes|license|lineage|macros|metrics|plugins|privacy_notice|profiling|project|py-modindex|scheduler|search|security|start|timezone|tutorial|ui)(\.html)?)$ "https://airflow.apache.org/docs/apache-airflow/stable/$1"
-RedirectMatch Permanent ^/docs/((1.10.1|1.10.2|1.10.3|1.10.4|1.10.5|1.10.6|1.10.7|1.10.8|1.10.9|1.10.10|1.10.11|1.10.12|1.10.13|stable)/.*)$ "https://airflow.apache.org/docs/apache-airflow/$1"
+RedirectMatch Permanent ^/docs/((1.10.1|1.10.2|1.10.3|1.10.4|1.10.5|1.10.6|1.10.7|1.10.8|1.10.9|1.10.10|1.10.11|1.10.12|1.10.13|1.10.14|2\.[0-9.]+|stable)/.*)$ "https://airflow.apache.org/docs/apache-airflow/$1"
```
mik-laj, line 6: I don't think we need redirects for all versions as we didn't publish these links before.

The only problem is `https://airflow.apache.org/docs/2.0.0/` which was included in the latest release by mistake. The following solution seems to me to address our problem better.
```
Redirect Permanent /docs/2.0.0/ https://airflow.apache.org/docs/apache-airflow/2.0.0/

File: landing-pages/site/static/.htaccess (hunk 2 above)
Comment (mik-laj) [reply landing-pages/site/static/.htaccess:4]: I don't think I need redirection for all files for 1.10.14 and 2.0.0 releases. We only need one redirection of one address for Airflow 2.0. See: 
https://github.com/apache/airflow-site/pull/362/files#r553838344

As for Airflow 1.10.14, we introduced a new link structure before this version was released.
Screenshot 2021-01-08 at 10 40 20

Comment (mik-laj) [reply landing-pages/site/static/.htaccess:4]: Ahh. 1.10.14 is also affected by this problem, so we can use the following requests.
```
RedirectMatch Permanent ^/docs/(1.10.14|2\.[0-9.]+)/$ "https://airflow.apache.org/docs/apache-airflow/$1"
```

mik-laj (approved):

PR #390, opened 2021-03-08
Add simple-dag-editor plugin

ohadmata:
Add simple dag editor plugin

### DIFF
--- a/landing-pages/site/content/en/ecosystem/_index.md
+++ b/landing-pages/site/content/en/ecosystem/_index.md
@@ -86,3 +86,5 @@ If you would you like to be included on this page, please reach out to the [Apac
 [Pylint-Airflow](https://github.com/BasPH/pylint-airflow) - A Pylint plugin for static code analysis on Airflow code.
 
 [whirl](https://github.com/godatadriven/whirl) - Fast iterative local development and testing of Apache Airflow workflows.
+
+[simple-dag-editor](https://github.com/ohadmata/simple-dag-editor) - Zero configuration Airflow plugin that let you manage your DAG files.


landing-pages/site/content/en/ecosystem/_index.md
```diff
@@ -86,3 +86,5 @@ If you would you like to be included on this page, please reach out to the [Apac
 [Pylint-Airflow](https://github.com/BasPH/pylint-airflow) - A Pylint plugin for static code analysis on Airflow code.
 
 [whirl](https://github.com/godatadriven/whirl) - Fast iterative local development and testing of Apache Airflow workflows.
+
+[simple-dag-editor](https://github.com/ohadmata/simple-dag-editor) - Zero configuration Airflow plugin that let you manage your DAG files.
```
turbaszek, line 90: Interesting one! Few questions:
- does it work with Airflow 2.0+?
- does it work with all executors?
- if someone is using synchronisation mechanism for syncing DAGs - which version will be used?

ohadmata: @turbaszek 
* I released new version today (0.1.0) that add a support for Airflow 2.0+
* The plugin just edit the DAG file itself, it's not related to the Airflow executors.
* The plugin is designed for managing dag files that mounted to the airflow-webserver at "dag_folder".
In the future I want to add S3 support, for git sync you have the code-editor plugin: 
https://pypi.org/project/airflow-code-editor

turbaszek: @ohadmata thanks for explanation!

> The plugin just edit the DAG file itself, it's not related to the Airflow executors.

It's important point to me. In most production deployments DAGs are either synchronised from external source (repository, bucket, etc). So my question is what will happen when users edit the DAG via the plugin? Will this change be overridden by some synchronisation proces?

> The plugin is designed for managing dag files that mounted to the airflow-webserver at "dag_folder".

Will this work with CeleryExecutor where webserver and workers are on separate nodes (pods, VMs etc) and do not share same volume (each node has it own volume which is synchronised with for example S3 bucket)?

As a user I miss answers to these questions. So I think it would be a good idea to add some explanation in documentation of your plugin so users are aware of some potential limitations or required setup of their deployment.

ohadmata: @turbaszek your remarks added to the documentation, thanks!

### Use-cases
* This plugin has been developed for a native deployment, where the Airflow webserver & scheduler are sharing the same dag_folder directory (as a mount).
* The plugin is not designed for deployments that synchronize the dag_folder from a remote source (Git/S3)
* You have the ability to disable the plugin for production env (see conf section)

turbaszek (approved):

PR #446, opened 2021-07-08
add link to airflow project sample

KimaruThagna:
Project sample is sufficiently commented from the readme to inline comments and hence can work as a resourceful tutorial.


potiuk:
I think, this is better to add in https://github.com/jghoman/awesome-apache-airflow

PR #492, opened 2021-10-05
Add Influx DB to list of providers

potiuk:


mik-laj (approved):
LGTM. 

Beside, we should add logos. There available in dev docs: http://apache-airflow-docs.s3-website.eu-central-1.amazonaws.com/

PR #533, opened 2021-12-31
Put copy button within code block

jmelot:
This is an attempt to address part of #324 (I'm not seeing the 404 that was also mentioned but happy to take a look at that as well if someone can point it out). 

It looks to me like the reason the copy button is in the correct place e.g. [here](https://airflow.apache.org/docs/apache-airflow/stable/tutorial.html) is that the `width: fit-content` on the pre elements is getting overridden by a `width: unset` from the `example-block-wrapper` class. So setting `width: 100%` appears to solve the problem, though there may be a better way to approach this.

### DIFF
--- a/landing-pages/site/assets/scss/_highlights.scss
+++ b/landing-pages/site/assets/scss/_highlights.scss
@@ -88,8 +88,7 @@ pre {
   padding: 16px 20px;
   border: solid 1px map-get($colors, very-light-pink);
   border-radius: 5px;
-  width: fit-content;
-  max-width: 100%;
+  width: 100%;
 
   span {
     @extend .monotext--brownish-grey;


potiuk:
But the copy button works for me: 

screenshot_copy

What's the browser you are using @jmelot

jmelot:
Hi @potiuk -- oops, I think my explanation of what I was trying to do was probably not very clear! The issue that was originally reported in #324 is that code blocks like the one [here](https://airflow.apache.org/docs/apache-airflow/stable/_modules/airflow/example_dags/tutorial.html) have the copy button outside the highlighted area, e.g.:

Screen Shot 2022-01-01 at 9 01 52 AM

You can see that the pre has width fit-content. In other pre blocks that have a parent element with the `example-block-wrapper` class, like the one you pointed out, the width property gets overridden:

Screen Shot 2021-12-31 at 4 40 15 PM

I see this in Firefox 95.0.2 and Chrome 96.0.4664.110

So that's why I just set `width: 100%`.

potiuk (approved):

PR #540, opened 2022-01-14
Only the table header should not wrap

malthe:
See https://github.com/readthedocs/sphinx_rtd_theme/issues/117 for a bigger discussion and many references.

### DIFF
--- a/landing-pages/site/assets/scss/_rst-content.scss
+++ b/landing-pages/site/assets/scss/_rst-content.scss
@@ -393,7 +393,7 @@
     margin-bottom: 0 !important
   }
 
-  .wy-table-responsive table td, .wy-table-responsive table th {
+  .wy-table-responsive table th {
     white-space: nowrap
   }
 


malthe:
An example of a documentation page where normal column wrapping would help is: https://airflow.apache.org/docs/apache-airflow-providers-databricks/stable/operators.html#databricksrunnowoperator.

There are probably some concerns for small devices, but not wrapping at all is not really a solution here either. That said, perhaps the whole theme should be updated to a newer version.

potiuk:
Did it work?

potiuk:
Can you post some screenshots here?

mik-laj:
This table should be replaced by a reference to the reference documentation that describes all the parameters for the operator. There is no point in repeating it in the guide again. The guide should describe the most common use cases and shows some examples for easy copy.
Here are reference docs for these operators: https://airflow.apache.org/docs/apache-airflow-providers-databricks/stable/_api/airflow/providers/databricks/operators/databricks/index.html

Moreover, this page lacks a page header, which disturbs the menu conventions used in all operators.
Screenshot 2022-01-23 at 01 01 27
Screenshot 2022-01-23 at 01 01 37

But going back to this contribution, it works for me.

airflow apache org_docs_apache-airflow-providers-databricks_stable_operators html

potiuk:
Is this one redy to merge ? WDYT @malthe @mik-laj ?

malthe:
@potiuk yes.

This is the before screenshot:

image

mik-laj (approved):

PR #572, opened 2022-04-05
Make the airflow sphinx theme as parallel safe

ashb:
This lets us run the sphinx-build with multiple workers  -- otherwise
sphinx will fallback to sequential mode.


potiuk:
Nice.

PR #575, opened 2022-04-22
Add warning to yanked cncf.kubernetes package documentation

potiuk:
The packages were yanked because of compatibility problems we
detected. This change just adds explicit warning message for all
the yanked packages.

Announcement here:
    
https://lists.apache.org/thread/fnccg9smgzbzorxvh4qqfj5l6w9y0y50

### DIFF
--- a/docs-archive/apache-airflow-providers-cncf-kubernetes/3.0.1/index.html
+++ b/docs-archive/apache-airflow-providers-cncf-kubernetes/3.0.1/index.html
@@ -612,12 +612,18 @@
 <div class="section" id="package-apache-airflow-providers-cncf-kubernetes">
 <h2>Package apache-airflow-providers-cncf-kubernetes<a class="headerlink" href="#package-apache-airflow-providers-cncf-kubernetes" title="Permalink to this headline">¶</a></h2>
 <p><a class="reference external" href="https://kubernetes.io/">Kubernetes</a></p>
-<p>Release: 3.0.1</p>
+<p>Release: 3.0.1:  <span class="pre" style="color: red">THIS VERSION IS YANKED - DO NOT INSTALL IT (SEE WARNING BELOW)</span></code></p>
 </div>
 <div class="section" id="provider-package">
 <h2>Provider package<a class="headerlink" href="#provider-package" title="Permalink to this headline">¶</a></h2>
 <p>This is a provider package for <code class="docutils literal notranslate"><span class="pre">cncf.kubernetes</span></code> provider. All classes for this provider package
 are in <code class="docutils literal notranslate"><span class="pre">airflow.providers.cncf.kubernetes</span></code> python package.</p>
+</div>
+<div class="admonition warning">
+<p class="admonition-title">Warning</p>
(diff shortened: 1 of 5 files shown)


potiuk:
image (6)

lwyszomi:
👍

docs-archive/apache-airflow-providers-cncf-kubernetes/3.0.1/index.html
```diff
@@ -612,13 +612,19 @@ <h2>Content<a class="headerlink" href="#content" title="Permalink to this headli
...
 are in <code class="docutils literal notranslate"><span class="pre">airflow.providers.cncf.kubernetes</span></code> python package.</p>
 </div>
+<div class="admonition warning">
+<p class="admonition-title">Warning</p>
+<p>This version of the provider has been yanked as it had compatibility issues. Do not use this version. If you are on Airflow 2.1.* or 2.2.*, please downgrade cncf.kubernetes to version 3.0.0</p>
+<p>The cncf.kubernetes provider is partially in the provider and partially in the Airflow core and this version of provider has dependencies on version of kubernetes library that clash with Airflow dependencies. In Airlfow 2.1.* and 2.2.* the latest working version of the provider is 3.0.0. The other 3.* versions of provider are "yanked" which means that they are skipped when installed unless you explicitly choose to install this version.</p>
+<p>Unless we see some serious bug to fix - all the future releases of "cncf.kubernetes" provider (starting from 4.0.0) will only be installable for Airflow 2.3.* and above.</p>
```
josh-fell, line 626:
```suggestion
<p>This version of the provider has been yanked as it had compatibility issues with core Airflow. Do not use this version. If you are on Airflow 2.1.* or 2.2.*, please downgrade the provider version to 3.0.0.</p>
<p>Both core Airflow and the cncf.kubernetes provider use the kubernetes library, however, the kubernetes version used for the provider clashes with core Airflow dependencies. In Airflow 2.1.* and 2.2.* the latest working version of the provider is 3.0.0. The other 3.* versions of provider are "yanked" which means that they are skipped when installed unless you explicitly choose to install this version.</p>
<p>Unless we see some serious bug to fix - all the future releases of the cncf.kubernetes provider (starting from 4.0.0) will only be installable for Airflow 2.3.* and above.</p>
```
Small typo fix and light wordsmithing. WDYT?

jedcunningham (approved):

jedcunningham:
Makes me wonder, should we have the warning for all the yanked versions on the list? e.g. in 3.1.2, have the same note for 3.1.1., 3.1.0, etc.

potiuk:
> Makes me wonder, should we have the warning for all the yanked versions on the list? e.g. in 3.1.2, have the same note for 3.1.1., 3.1.0, etc.

We do actually :P

potiuk:
image

jedcunningham:
I more meant, 4.0.0 has no mention that 3.1.2-3.0.1 are yanked. If you go to 3.1.2 and hit 3.1.1 in the right nav you see no yank warning.

potiuk:
Ah yeah. I seee - in the Changelog. Right. Should be added too. Looks like lots of doc updates today :D

potiuk:
Added in https://github.com/apache/airflow-site/pull/580

PR #576, opened 2022-04-22
Migrate Google Analytics to Matomo from the ASF

rossturk:
Hello! This replaces the Google Analytics tracking tag in the airflow.apache.org website with a Matomo one from the ASF. I propose this because the ASF has recently updated its privacy policy, which now prohibits use of Google Analytics.

There are three changes in this PR:
1. Disabling GA in the Hugo config
2. Replacing the tag in the Sphinx theme
3. Replacing the tag in all of the prebuilt `docs-archive` content (about 80k pages!)

For #3, I used this script:
https://gist.github.com/rossturk/1223ae5d57fbcbb4bc32da0d49137ef2

PR #577, opened 2022-04-22
Add Matomo tag in the proper location for Docsy

rossturk:
This is a followup to #576.

I didn't make the change in the right place for `landing-pages`, and I have now done so. I've also verified that the contents of `dist/` in my local build now contain the desired tags.


potiuk (approved):
This is Cool. I already see sesssions appearing in https://analytics.apache.org/

And the cool thing is - this is public info !

PR #579, opened 2022-04-25
Bump theme to switch to Matomo

potiuk:


potiuk:
cc: @rossturk

rossturk:
Ah, I see. As I understand it, this will cause new doc builds to get the updated tracking tag 👍

jedcunningham (approved):

potiuk:
> Ah, I see. As I understand it, this will cause new doc builds to get the updated tracking tag 👍

Looks like after consulting with Kamil - I need to publish it to PypI first though

PR #581, opened 2022-04-26
Use Matomo for page feedback mechanism

rossturk:
I was made aware that the page feedback mechanism (i.e., "do you find this page helpful?") in each docs page also uses Google Analytics. As a followup to #576, this moves the mechanism over to Matomo.

Each time a user clicks on the rating button, it will send an event to Matomo. Events in Matomo have a `category` and an `action`, along with an optional `name` and `value`. I have chosen Docs for the category and Rating as the action - the name is set to the current page path and the value is the numerical rating.

This required a lot of individual file changes in `docs-archive`. Sorry about the big PR, again! I think it's safe to ignore commit e933c52 :)

Unfortunately, I haven't yet been able to cleanly fix the feedback mechanism for the landing pages. The feature is currently implemented in Docsy, and somewhat tied to GA. For now it has been disabled in this PR.

(This also, as a side effect, more completely deploys the fix in #534. While comparing all of the various `docs.js` files in the docs archive, I found that they were all the same except for some where that fix had not yet been applied.)

Finally, this bumps the version of the Sphinx theme, so that future doc builds will include the updated mechanism. I'm not sure if the PyPI publication is automatic or not :)


kaxil (approved):

potiuk:
I just uploaded the PyPI package and restarted the doc build in ~ 20 minutes the latest changes should be live at our page - so worth checking it :)

rossturk:
@potiuk It looks correct to me 👍

potiuk:
Cool! thanks!

PR #594, opened 2022-05-16
Add Airflow Summit CTA.

potiuk:


potiuk:
image (10)

potiuk:
Good enough :)

ryanahamilton (approved):

potiuk:
@mik-laj - not sure - do we have to publish a theme after that ( I never know :) )

kaxil:
> @mik-laj - not sure - do we have to publish a theme after that ( I never know :) )

If we want to release a new version of Airflow, providers or Helm Chart, then yes -- or else we will need to apply these changes manually. So not 100% needed per see if we want to find/replace manually as we will revert it after the Summit

potiuk:
We want to submit providers docs now :) . So yeah

potiuk:
Will do it then :)

PR #610, opened 2022-06-09
Update links to example dags to point to the right github URL

potiuk:
The links to example DAGs from all versions of the providers
pointed to "main" version of the example DAGS. This was wrong
because it always shown the "latest" version of the examples,
and it got additionally broken by AIP-47 where example_dags
folder is essentially removed from all providers.

This change updates the links to the "right" version of
example dags for all providers, including cleaning up
and removing the links that were pointing to non-existing
example_dags folders (because example_dags were missing in
the specific version of the provider).

This is related to https://github.com/apache/airflow/issues/24331

### DIFF
--- a/docs-archive/apache-airflow-providers-airbyte/1.0.0/index.html
+++ b/docs-archive/apache-airflow-providers-airbyte/1.0.0/index.html
@@ -331,7 +331,6 @@
 </ul>
 <p class="caption"><span class="caption-text">Resources</span></p>
 <ul>
-<li class="toctree-l1"><a class="reference external" href="https://github.com/apache/airflow/tree/master/airflow/providers/airbyte/example_dags">Example DAGs</a></li>
 <li class="toctree-l1"><a class="reference external" href="https://pypi.org/project/apache-airflow-providers-airbyte/">PyPI Repository</a></li>
 </ul>
 <p class="caption"><span class="caption-text">Commits</span></p>
@@ -476,7 +475,6 @@
 </ul>
 <p class="caption"><span class="caption-text">Resources</span></p>
 <ul>
-<li class="toctree-l1"><a class="reference external" href="https://github.com/apache/airflow/tree/master/airflow/providers/airbyte/example_dags">Example DAGs</a></li>
 <li class="toctree-l1"><a class="reference external" href="https://pypi.org/project/apache-airflow-providers-airbyte/">PyPI Repository</a></li>
 </ul>
 <p class="caption"><span class="caption-text">Commits</span></p>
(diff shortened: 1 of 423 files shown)


potiuk:
cc: @deeptansh

kaxil (approved):

docs-archive/apache-airflow-providers-airbyte/1.0.0/index.html
```diff
@@ -331,7 +331,6 @@
 </ul>
 <p class="caption"><span class="caption-text">Resources</span></p>
 <ul>
-<li class="toctree-l1"><a class="reference external" href="https://github.com/apache/airflow/tree/master/airflow/providers/airbyte/example_dags">Example DAGs</a></li>
```
josh-fell, line 334: Any idea what the Airbyte example DAGs get removed in v1.0.0, appear in v2.0.0 (pointing the expected tag), disappear again in v2.1.0, and then reappear in v2.1.2?

docs-archive/apache-airflow-providers-apache-drill/1.0.0/index.html
```diff
@@ -330,7 +330,6 @@
 </ul>
 <p class="caption"><span class="caption-text">Resources</span></p>
 <ul>
-<li class="toctree-l1"><a class="reference external" href="https://github.com/apache/airflow/tree/main/airflow/providers/apache/drill/example_dags">Example DAGs</a></li>
```
josh-fell, line 333: All of the references to Apache Drill DAGs disappear?

potiuk: Yes. It never had any examples :). So the links were just wrong.

josh-fell: But it does for that release: https://github.com/apache/airflow/tree/providers-apache-drill/1.0.0/airflow/providers/apache/drill/example_dags

Granted it didn't have an `__init__.py` file in that directory for 1.0.0, but I assume that doesn't matter?

josh-fell: Same pattern with the newest provider, ArangoDB: https://github.com/apache/airflow/tree/providers-arangodb/1.0.0/airflow/providers/arangodb/example_dags

docs-archive/apache-airflow-providers-airbyte/1.0.0/index.html
potiuk: I think it was because of some intermittent issue. I was checking if the URL is reachable and likely there was a temporary issue with GitHub at this moment. I did watch it while running so even if this happened, that would not be a "widesprad" issue. I would not worry too much about finding and fixing those issues though - at the worst case soneone does not find a link to a specific version's examples which is better than having a link that leads to "Missing".

I think it could have been done with only actually reacting to 404 (maybe this case was like 500. But it's a done deal now :)

PR #611, opened 2022-06-09
Fix mistakenly removed links in Providers

potiuk:
The #610 removed too many links because GitHub throttled checking
if URLS are available.

This PR is re-done with better checking for error code.


josh-fell (approved):

kaxil:
Good eyes @josh-fell

Issue #338, opened 2020-12-09
Rewrite ./site.sh to Python


mik-laj:
Dear friends.

In this project, we have one bash script - [./site.sh](https://github.com/apache/airflow-site/blob/master/site.sh) that starts to grow more and more, so its maintenance becomes more problematic. I think it's a good idea to rewrite it into Python.

However, I would like us to continue limiting the use of external libraries. For Docker interactions, it is better to use docker-cli than the Docker library.

We use subcommands in the script. I am attaching an example of how to do this in Python.
https://gist.github.com/mik-laj/ff008718fc6cec9fe929731b8c62d6f8

Best regards,
Kamil Breguła

CC: @brennoflavio

mik-laj:
Assigned.

pankajastro:
Hi @jedcunningham can I take this? thank you!

pierrejeambrun:
@pankajastro just assigned you :)

pierrejeambrun:
I believe the issue is still valid, go ahead, I just assigned you.

NarmalaSk:
I have kept a pr [](https://github.com/apache/airflow-site/pull/1161) and also we can go for development of the airflow site using the getpelican as mentioned here I have interest in this issue [](https://github.com/apache/airflow-site/issues/719)

NarmalaSk:
I have kept a pr everything with new script works fine but I have ran into some errors in ci for site.py I have cleared them but now got some errors in precommit files that are regarding some spaces, Also build time is so long I want to enhance that also I will keep a pr for that also

NarmalaSk:
Hey my pull request has been closed it's ok with that, But just let me know after migration versions to s3 done, I think my efforts for migrating script to python wouldn't go vague

potiuk:
It's done :)

NarmalaSk:
> It's done :)

Should I keep a pr for site.py again or how the process would be @potiuk

NarmalaSk:
Ok I'll start from the scratch of new script, Now airflow-site repo is added with new code right as you said, yeah potiuk then what os the status with replacement of hugo+npm

NarmalaSk:
I dont Know this is possible or not ,I will work on both issues of conversion of npm+hugo+Bash, Python script site.py. Does Apache airflow  participate in LFX Mentorship if yes If Apache foundation introduce this Both issues as a project .It would help me more ,Its ok if its not possible @potiuk what do you think ??

kaxil:
[...]

PR #623, opened 2022-07-04
Remove Bash from the process of building the site

Bowrna:
closes: #338

### DIFF
--- a/dump_docs_packages_metadata.py
+++ b/dump_docs_packages_metadata.py
@@ -1,0 +1,40 @@
+import json
+import sys
+
+import os
+from typing import List
+import semver
+
+
+ROOT_DIR = os.path.dirname(os.path.join(os.path.dirname(os.path.abspath(__file__)), os.path.pardir))
+APACHE_AIRFLOW_ARCHIVE = os.path.join(ROOT_DIR, "docs-archive")
+
+
+def get_all_versions(directory: str) -> List[str]:
+    return sorted(
+        (f for f in os.listdir(directory) if os.path.isdir(os.path.join(directory, f))),
+        key=lambda d: semver.VersionInfo.parse(d),
+    )
+
+
+def get_stable_version(directory: str):
+    with open(os.path.join(directory, "stable.txt")) as f:
+        return f.read().strip()
+
+
+def dump_docs_package_metadata(dump_file_path):
+    all_packages_infos = [
+        {
+            "package-name": package_name,
+            "stable-version": get_stable_version(os.path.join(APACHE_AIRFLOW_ARCHIVE, package_name)),
+            "all-versions": get_all_versions(os.path.join(APACHE_AIRFLOW_ARCHIVE, package_name)),
+        }
+        for package_name in os.listdir(APACHE_AIRFLOW_ARCHIVE)
+        if (
+            not package_name.startswith(".") and  # Exclude .DS_Store/
+            os.path.isfile(os.path.join(os.path.join(APACHE_AIRFLOW_ARCHIVE, package_name, 'stable.txt')))
+        )
+    ]
+    dump_file = open(dump_file_path, 'w')
+    json.dump(all_packages_infos, dump_file, indent=2)
+    dump_file.close()
--- a/site_tool.py
+++ b/site_tool.py
@@ -1,0 +1,258 @@
+import argparse
+import contextlib
+from pathlib import Path
+import os
+import glob
+import re
+import sys
+import shutil
+import subprocess
+from typing import Union, List, Optional, Mapping
+from dump_docs_packages_metadata import dump_docs_package_metadata
+# from bs4 import BeautifulSoup
+IMAGE_NAME="airflow-site"
+CONTAINER_NAME="airflow-site-c"
+RunCommandResult = Union[subprocess.CompletedProcess, subprocess.CalledProcessError]
+
+
+def run_command(cmd: List[str],
+    env: Optional[Mapping[str, str]] = None,
+    cwd: Optional[Path] = None,
+    input: Optional[str] = None,
+    check: bool = True,
+     **kwargs,
+)-> RunCommandResult:
+    workdir: str = str(cwd) if cwd else os.getcwd()
+    try:
+        cmd_env = os.environ.copy()
+        cmd_env.setdefault("HOME", str(Path.home()))
+        if env:
+            cmd_env.update(env)
+        return subprocess.run(cmd, input=input, check=check, env=cmd_env, cwd=workdir, **kwargs)
+    except subprocess.CalledProcessError as ex:
+        if ex.stdout:
+            print(ex.stdout)
+        if ex.stderr:
+            print(ex.stderr)
+        return ex
+
+def prepare_package_metadata(file_path):
+    print("Preparing packages-metadata.json")
+    dump_docs_package_metadata(file_path)
+
+def sanity_checks_files(file_list):
+    for file_to_check in file_list:
+        if not file_to_check.exists():
+            print("Missing file:", str(file_to_check))
+            sys.exit(1)
+
+def build_site():
+    print("Building full site")
+    landing_page = Path("landing-pages/dist/index.html")
+    if not landing_page.exists():
+        build_landing_page()
+    DIST = Path.cwd()/ "dist"
+    # DIST.mkdir(parents=True, exist_ok=True)
+    # files = DIST.glob('**/*')
+    # for file_to_delete in files:
+    shutil.rmtree(str(DIST))
+    shutil.copytree("landing-pages/dist/.","dist/")
+    for package_name in next(os.walk('docs-archive'))[1]:
+        stable_path = Path('docs-archive')/ package_name/ 'stable.txt'
+        if stable_path.exists():
+            for version in next(os.walk('docs-archive/'+package_name))[1]:
+                shutil.copytree(str(Path("docs-archive")/package_name/version/"."), str(Path("dist")/"docs"/package_name/version))
+            stable_version = stable_path.read_text()
+            shutil.copytree(str(Path("docs-archive")/package_name/stable_version/"."), str(Path("dist")/"docs"/package_name/"stable"))
+        else:
+            shutil.copytree("docs-archive/"+package_name+"/.", "dist/docs/"+package_name)
+    prepare_package_metadata(Path.cwd()/"dist"/"_gen"/"packages-metadata.json")
+    files_to_check = [
+        Path.cwd()/Path("dist/docs/index.html"),
+        Path.cwd()/Path("dist/docs/apache-airflow/1.10.7/tutorial.html"),
+        Path.cwd()/Path("dist/docs/apache-airflow/stable/tutorial.html"),
+        Path.cwd()/Path("dist/_gen/packages-metadata.json"),
+    ]
+    sanity_checks_files(files_to_check)
+
+def run_build_site(args):
+    """
+    Prepare dist directory with landing pages and documentation.
+    """
+    ensure_node_module_exists()
+    build_site()
+
+def ensure_node_module_exists():
+    print("Checking if node module exists")
+    node_dependency = Path(Path.cwd()/'landing-pages'/'node_modules')
+    if not node_dependency.exists():
+        print("Missing node dependencies. Start installation.")
+        run_command(['yarn', 'install'], cwd=Path.cwd()/'landing-pages')
+        print("Dependencies installed.")
+
+def ensure_that_website_is_build():
+    print("Check if landing-pages/dist/index.html file exists")
+    landing_page = Path.cwd()/Path("landing-pages/dist/index.html")
+    if not landing_page.exists():
+        print("The website is not built. Start building.")
+        run_command(['npm', 'run', 'build'], cwd=Path.cwd()/"landing-pages")
+        print("The website is built.")
+
+def run_install_node_deps(args):
+    """
+    Download all the Node dependencies.
+    """
+    run_command(['yarn', 'install'], cwd=Path.cwd()/"landing-pages")
+
+def preview_landing_pages():
+    run_command(['yarn', 'run', 'index'], cwd=Path.cwd()/"landing-pages")
+    prepare_package_metadata(Path.cwd()/"landing-pages/site/static/_gen/packages-metadata.json")
+    run_command(['yarn', 'run', 'preview'], cwd=Path.cwd()/"landing-pages")
+
+def run_preview_landing_pages(args):
+    """Starts the web server with preview of the website."""
+    ensure_node_module_exists()
+    preview_landing_pages()
+
+def build_landing_page():
+    print("Building landing page")
+    run_command(['yarn', 'run', 'index'], cwd=Path.cwd()/"landing-pages")
+    prepare_package_metadata(Path.cwd()/"landing-pages/site/static/_gen/packages-metadata.json")
+    run_command(['yarn', 'run', 'build'], cwd=Path.cwd()/"landing-pages")
+
+def run_build_landing_pages(args):
+    """Build the landing pages."""
+    ensure_node_module_exists()
+    build_landing_page()
+
+@contextlib.contextmanager
+def working_directory(source_path: Path):
+    prev_cwd = Path.cwd()
+    os.chdir(source_path)
+    try:
+        yield
+    finally:
+        os.chdir(prev_cwd)
+
+# def extract_links(file_path):
+#     links = []
+#     with open(file_path) as file_content:
+#         urls = file_content.read()
+#         #  grep "^https\?://"
+#         links = re.findall(r'(https?://[^\s]+)', urls)
+#         # links = re.findall( "^https\?://", urls)
+#     # print(links)
+#     return links
+
+# @working_directory(Path.cwd()/"landing-pages")
+# def run_check_site_links(args):
+#     dist = Path.cwd()/"dist"
+#     index_page = dist/"index.html"
+#     if not index_page.exists():
+#         print("You should build website first.")
+#         sys.exit(1)
+#     html_files = dist.glob('**/*.html')
+#     count = 0
+#     external_links = set()
+#     for html_file in html_files:
+#         # print(html_file)
+#         count += 1
+#         external_links.update(extract_links(html_file))
+#     with open("test.txt","w") as test:
+#         for item in external_links:
+#             test.write("%s\n" % item)
+#     print(external_links, len(external_links))
+
+def run_check_site_links(args):
+    """Check to make sure that links are correct in the website."""
+    ensure_node_module_exists()
+    ensure_that_website_is_build()
+    run_command(['./check-links.sh'], cwd=Path.cwd()/"landing-pages")
+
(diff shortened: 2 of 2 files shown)


site.py
```diff
@@ -0,0 +1,233 @@
...
+        cmd_env.setdefault("HOME", str(Path.home()))
+        if env:
+            cmd_env.update(env)
+        return subprocess.run(cmd, input=input, check=check, env=cmd_env, cwd=workdir, **kwargs)
+    except subprocess.CalledProcessError as ex:
+        if ex.stdout:
+            print(ex.stdout)
+        if ex.stderr:
+            print(ex.stderr)
+        return ex
```
turbaszek, line 34: Do we need to return an error here? Would it be better to raise it?

site.py
```diff
@@ -0,0 +1,233 @@
...
+
+
+def check_docker_environment():
+    is_docker_env = False
+    docker_env_file = Path(".dockerenv")
+    if docker_env_file.exists():
+        is_docker_env = True
+    return is_docker_env
```
turbaszek, line 42:
```suggestion

    return docker_env_file.exists()
```

site.py
Bowrna: This is a common method to execute the script via Python subprocess. If I could return the error, then I can log the error message according to the command executed. If you have any other options please let me know @turbaszek

Bowrna:
https://github.com/apache/airflow-site/blob/988ae480223f723efb0b2f97992d30f42128a0ef/site.sh#L77-L114

Can someone explain to me what's happening in this code part? i find it difficult to understand what's happening here.

@mik-laj could you help me?

mik-laj:
@Bowrna see: https://stackoverflow.com/a/12498485   This generates a relative path for use in a container.

Bowrna:
> @Bowrna see: https://stackoverflow.com/a/12498485 This generates a relative path for use in a container.

Could you tell me why we need to convert the absolute path into relative path to run the script?  Wouldn't it run fine with the absolute path? @mik-laj

potiuk:
Because in container it will be in a different location ("/opt/airlfow/.....") rather than `<YOUR LOCAL AIRLFOW SOURCES>`/....

Bowrna:
Ah I see @potiuk. As the sh file is replaced to run without docker dependency, i forgot about this point. Thanks

Bowrna:
https://github.com/apache/airflow-site/blob/fd7af05318e34f4efd6b8446ba8729112f30ee08/site.sh#L199

Is this redirect still valid even after removing the docker part @mik-laj @potiuk @turbaszek @rossturk 
Could anyone help me to understand?

I have this doubt because we are redirecting to the path "/docs/${package_name}/stable/index.html". So I wanted to ensure if this is still valid even after removing the docker part for building site. Thanks

potiuk:
> Is this redirect still valid even after removing the docker part @mik-laj @potiuk @turbaszek @rossturk

Nope. Not needed. I think it was not even needed before for a long time already. This is likely a remnant of a very old approach - long time defunct.

I think what is important is that after buildng the site locallly, you can start the server and locally browse the site, also a check that after deploying to airflow site it still works should happen, but this is something we can likely do only after we merge it.

Bowrna:
@potiuk I get the following error when building the site and I am not sure how I can solve this. This occurs when `hugo` command runs under the hood.  This happens when running the existing `site.sh` script too


```
Start building sites …
hugo v0.101.0+extended darwin/arm64 BuildDate=unknown
INFO 2022/07/26 19:15:23 syncing static files to /
WARN 2022/07/26 19:15:23 found no layout file for "HTML" for layout "search" for kind "page": You should create a template file which matches Hugo Layouts Lookup Rules for this combination.
WARN 2022/07/26 19:15:23 found no layout file for "HTML" for kind "section": You should create a template file which matches Hugo Layouts Lookup Rules for this combination.
WARN 2022/07/26 19:15:23 found no layout file for "HTML" for kind "section": You should create a template file which matches Hugo Layouts Lookup Rules for this combination.
WARN 2022/07/26 19:15:23 found no layout file for "HTML" for kind "taxonomy": You should create a template file which matches Hugo Layouts Lookup Rules for this combination.
WARN 2022/07/26 19:15:23 found no layout file for "HTML" for kind "taxonomy": You should create a template file which matches Hugo Layouts Lookup Rules for this combination.
ERROR 2022/07/26 19:15:23 render of "page" failed: "/Users/sathishkannan/code/airflow-site/landing-pages/site/layouts/blog/baseof.html:23:7": execute of template failed: template: blog/single.html:23:7: executing "blog/single.html" at <partial "head.html" .>: error calling partial: partial "head.html" not found
ERROR 2022/07/26 19:15:23 render of "page" failed: "/Users/sathishkannan/code/airflow-site/landing-pages/site/layouts/blog/baseof.html:23:7": execute of template failed: template: blog/single.html:23:7: executing "blog/single.html" at <partial "head.html" .>: error calling partial: partial "head.html" not found
ERROR 2022/07/26 19:15:23 render of "page" failed: "/Users/sathishkannan/code/airflow-site/landing-pages/site/layouts/blog/baseof.html:23:7": execute of template failed: template: blog/single.html:23:7: executing "blog/single.html" at <partial "head.html" .>: error calling partial: partial "head.html" not found
ERROR 2022/07/26 19:15:23 render of "page" failed: "/Users/sathishkannan/code/airflow-site/landing-pages/site/layouts/blog/baseof.html:23:7": execute of template failed: template: blog/single.html:23:7: executing "blog/single.html" at <partial "head.html" .>: error calling partial: partial "head.html" not found
Error: Error building site: failed to render pages: render of "page" failed: "/Users/sathishkannan/code/airflow-site/landing-pages/site/layouts/blog/baseof.html:23:7": execute of template failed: template: blog/single.html:23:7: executing "blog/single.html" at <partial "head.html" .>: error calling partial: partial "head.html" not found
Total in 825 ms
error Command failed with exit code 255.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
error Command failed with exit code 255.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

rossturk:
> Error: Error building site: failed to render pages: render of "page" failed: "/Users/sathishkannan/code/airflow-site/landing-pages/site/layouts/blog/baseof.html:23:7": execute of template failed: template: blog/single.html:23:7: executing "blog/single.html" at <partial "head.html" .>: error calling partial: partial "head.html" not found

Hmmmm I may be wrong here but if I remember correctly this is one of the errors that occurs when the submodules aren't checked out.

Have you run `git submodules update --init`?

Bowrna:
> Have you run `git submodules update --init`?

Yes after running this the `build-site` command works. Thanks, @rossturk. 
I have error occurring in `check-site-links` command and it fails with following error message.

```
sathishkannan@Sathishs-MacBook-Air airflow-site % python site_tool.py check-site-links
Checking if node module exists
Check if landing-pages/dist/index.html file exists
/opt/homebrew/bin/lynx
Working directory: /Users/sathishkannan/code/airflow-site/landing-pages
Dist directory: ./dist/

./check-links.sh: line 61: readarray: command not found
sathishkannan@Sathishs-MacBook-Air airflow-site % ./site.sh check-site-links
2022-07-27 11:32:33:INFO: Checking if node module exists
2022-07-27 11:32:33:INFO: Check if /Users/sathishkannan/code/airflow-site/landing-pages/dist/index.html file exists
2022-07-27 11:32:33:INFO: Changing to /Users/sathishkannan/code/airflow-site/landing-pages/ and running: ./check-links.sh@
/opt/homebrew/bin/lynx
Working directory: /Users/sathishkannan/code/airflow-site/landing-pages
Dist directory: ./dist/

./check-links.sh: line 61: readarray: command not found
sathishkannan@Sathishs-MacBook-Air airflow-site %
```
Am I missing downloading anything here? I checked in CONTRIBUTORS.md guide but didn't find anything relevant to this. 
I am working on Mac M1

rossturk:
> ./check-links.sh: line 61: readarray: command not found

@Bowrna I think this has to do with the version of bash. If I try this on my M1 Mac, I get an error:

```bash
bash-3.2$ bash -version
bash -version
GNU bash, version 3.2.57(1)-release (arm64-apple-darwin21)
Copyright (C) 2007 Free Software Foundation, Inc.

bash-3.2$ readarray foo < <(ls -al)
readarray foo < <(ls -al)
ls -al
bash: readarray: command not found
```

But if I try it on my Linux machine, I get this:

```bash
rturk@caymus:~$ bash --version
GNU bash, version 5.0.17(1)-release (x86_64-pc-linux-gnu)
Copyright (C) 2019 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

rturk@caymus:~$ readarray foo < <(ls -al)

rturk@caymus:~$ echo ${foo[8]}
drwxr-xr-x 2 rturk caymus 4096 Oct 6 2021 .dbt
```

I think perhaps the `CONTRIBUTING.md` should instruct folks to install bash from Homebrew. I haven't tested it myself, but this looks promising:

```
% brew info bash
bash: stable 5.1.16 (bottled), HEAD
...
```

potiuk:
I think it has never been supposed to be run on Mac :). 

And one of the reasons we want to rewrite everything in Python, is that bash is no longer "common" accross multiple platform. The MacOS Bash is Bash 3 (for licencing reason - Apple refused to use Bash4 due to GPLv3 (https://news.ycombinator.com/item?id=20102597)  and a lot of tools there are not really POSIX compliant, so there are plenty of hacks around that (readarray is one of the features of Bash4).

The ultiimate goal we have is that Breeze and all our tools can be used WITHOUT having or using bash in the Host. Fully, Once we get there, the next step will be to make sure that it works on Windows.

And that should be our goal here as well (even though site.sh is not really used by many people). So - let's just not mention in CONTRIBUTING.md that you need to install bash from brew. I deliberately avoided that necessity, because that it yet another prerequisite that adds friction to the first-time experience. 

Ideally just `Python3 + Docker + Docker-Compose` should be enough to do everything developer needs to do in Airlfow.

potiuk:
> I think perhaps CONTRIBUTING.md should instruct folks to install bash from Homebrew. I haven't tested it myself, but this looks promising:

This has profound consequences. Many tools which "assume" bash is v3 on Mac will stop working if you make Bash 4 default. Even if you try to install it via `brew` it will tell you that it is 'cask-only`, which means that it is not becoming a default `bash` - you need to perform an extra step to make it default precisely because it might break a lot of things on Mac.  

Going bash-less is the only good route we can take - now when we have Python3.7+ everywhere (previously various options of having Python2 + Python 3 (3.5 as well)  created similar set of problems, but since everyone is now on Python 3.7+ (3.6 EOL reached) basing "lowest common denominator" on Python 3.7 is great for such an environment.

rossturk:
> > I think perhaps CONTRIBUTING.md should instruct folks to install bash from Homebrew. I haven't tested it myself, but this looks promising:
> 
> 
> 
> This has profound consequences. Many tools which "assume" bash is v3 on Mac will stop working if you make Bash 4 default. Even if you try to install it via `brew` it will tell you that it is `cask-only`, which means that it is not becoming a default `bash` - you need to perform an extra step to make it default precisely because it might break a lot of things on Mac.  

I'm sorry, what are you suggesting as an alternative?

potiuk:
> I'm sorry, what are you suggesting as an alternative?

Remove Bash. Completely. I think this whole task is about removing bash and replacing it with Python :).
I thin the problem @Bowrna has (correct me @Bowrna if I am wrong) that you cannot really run the OLD script corrrectly. This can be fixed locally by updating bash with brew and setting temporary PATH to get it first on path. Once the script is replaced with Bash, there is no need to mention it in  CONTRIBUTING - because ..... there will be no bash to run any more.

Or am I missing something :)?

rossturk:
> > I'm sorry, what are you suggesting as an alternative?
> 
> Remove Bash. Completely. I think this whole task is about removing bash and replacing it with Python :). I think the problem @Bowrna has (correct me @Bowrna if I am wrong) that you cannot really run the OLD script corrrectly. This can be fixed locally by updating bash with brew and setting temporary PATH to get it first on path. Once the script is replaced with Bash, there is no need to mention it in CONTRIBUTING - because ..... there will be no bash to run any more.
> 
> Or am I missing something :)?

Yes, you are missing that this PR title is "replace site.sh" and the script we are talking about is `landing-pages/check-links.sh`. Are you suggesting that the scope of this PR is expanded to include *all* shell scripts?

potiuk:
> Yes, you are missing that this PR title is "replace site.sh" and the script we are talking about is landing-pages/check-links.sh. Are you suggesting that the scope of this PR is expanded to include all shell scripts?

Ahhhh. I see now :facepalm: . 

Precisely. This was the whole purpose of the task - to replace all the bash usage. It makes no sense only to replace site.sh. The reason why we are doing it is not because "we do not like bash", but because "we want to get rid of bash from the host entirely - to be able to run it all on MacOS, Linux, Windows or anything else that has "native" Python (and potentially has bash).

So the TITLE of the task is wrong. Should be "Remove Bash from the process of building the site".

potiuk:
And the root cause of the confusion was that when the original issue was created, there was indeed only one `site.sh` :

See the description of the issue by @mik-laj - so that's why the title was wrong 

https://github.com/apache/airflow-site/issues/338

> In this project, we have one bash script - [./site.sh](https://github.com/apache/airflow-site/blob/master/site.sh) that starts to grow more and more, so its maintenance becomes more problematic. I think it's a good idea to rewrite it into Python.

Bowrna:
Ah, I see the whole point now. Currently to see how this `check-site-links` command, I have to install bash via brew and see what is happening under the hood. I will try converting that part to python after understanding what is happening in that part of code.

Bowrna:
Trying to understand the below bash command. Could anyone tell me what's happening in this command
https://github.com/apache/airflow-site/blob/445586c1552918e9531cd13d9cd64c0e63f5da03/landing-pages/check-links.sh#L59-L63
When trying to execute the above script I just enabled print and saw it's logging a few html files.
Screenshot 2022-07-28 at 11 48 17 PM

But when I took out that command and executed it via terminal it shows a big list of HTML files. I don't understand how this is different from executing in a script. When executing in terminal I see `index.html` from all providers and its different version say example `/dist//docs/apache-airflow-providers-imap/1.0.0/_modules/index.html`. But i don't see them when executing it via `check-links.sh`
Screenshot 2022-07-29 at 12 04 18 AM

potiuk:
It fills the array 'pages' with the list of HTML files found. The problem you see is likely the `-print 0` flag- rather than producing files separated by "spaces" it separates them with \0 (null character) so that there is no confusion between space being separator between files and one that is part of the file name.

so if you have 

* `a file.html`
* `another file.html`

This wil produce

"a file.html\0another file.html"

Then when you read it with -d '' you will get array:

['a file.html', 'another file.html'] 

Otherwise you'd get:

['a', 'file.html', 'another', 'file.html']

The problem with differences you see is that the "\0" separated files are not broken over multiple lines because they are really a very long line with no spaces to allow the break happen.

Bowrna:
> When executing in terminal I see `index.html` from all providers and its different version say example `/dist//docs/apache-airflow-providers-imap/1.0.0/_modules/index.html`. But i don't see them when executing it via `check-links.sh`

Could you share your view for this point? This is also another difference in the output when executing it in different ways @potiuk .

potiuk:
> > When executing in terminal I see `index.html` from all providers and its different version say example `/dist//docs/apache-airflow-providers-imap/1.0.0/_modules/index.html`. But i don't see them when executing it via `check-links.sh`
> 
> Could you share your view for this point? This is also another difference in the output when executing it in different ways @potiuk .

I think it's just the matter of when it is called and whether the files are there when it is executed. Some earlier steps likely prepare the files in dist folder and maybe that step is missing when you run it manually ? Can't think of any other reason.

Bowrna:
I have tried executing both of ways consecutively without running any other step in between. Let me debug this again and come with more information @potiuk

potiuk:
Hey @Bowrna are you going to continue with that one :) ? Not the highest prirority, so if we do not plan it, maybe better to close that one?

Bowrna:
I will work on this part this weekend @potiuk. As I got stuck on resolving the issue, I started fixing other issues and slowly I became dormant in working on this issue.

mik-laj:
> https://github.com/apache/airflow-site/blob/fd7af05318e34f4efd6b8446ba8729112f30ee08/site.sh#L199
> 
> Is this redirect still valid even after removing the docker part @mik-laj @potiuk @turbaszek @rossturk Could anyone help me to understand?

It is common to delete part of a URL to go to the parent location.
For example, when a user visits a URL:
```
https://airflow.apache.org/docs/apache-airflow-providers-alibaba/2.1.0/index.html
```
then the user can delete part of the URL and go to:
```
https://airflow.apache.org/docs/apache-airflow-providers-alibaba/
```
When we do not have this redirect, the user will see a 404 error.

In other words, all directories the user has access to should have an index.html file to improve UX.

potiuk:
Approved to see.

eladkal:
@Bowrna are you still working on this PR?

Bowrna:
@eladkal I didn't work on this for a long time as I was working on other PR. I will start working on this PR and one another in this weekend as I have some time due to the upcoming holiday week.

Bowrna:
As I am currently working on this PR, I will post the parts that I have completed and parts that are pending to keep on track.
Following are the commands supported in the `site.sh` bash script that's being moved to python:
1. install-node-deps (successfully moved to python as of now)
2. preview-landing-pages (successfully moved to python as of now)
3. build-site (successfully moved to python as of now)
4. build-landing-pages (successfully moved to python as of now)
5. check-site-links (only pending part)
6. prepare-theme (successfully moved to python as of now)
7. lint-js (successfully moved to python as of now)
8. lint-css (successfully moved to python as of now)

Bowrna:
@potiuk @mik-laj I am considering to use beautiful soup library, to extract the links from html files for check-site-links command. Is it fine or is there any better way to handle it? If there is please let me know. 

Thanks. :) 
I hope to finish this before this year end 🤞

mik-laj:
We don't use this check, so I think we can drop it for now.

Bowrna:
@mik-laj Could you tell me if the check-site-links command in site.sh is used or not? Because that command uses this script file and here they have used lynx to extract URL? 

Could you give me more explanation on that part?

Bowrna:
Also, can this code be reviewed? I will try fixing this PR and getting it merged to main.

Bowrna:
Hello all :) do I have any update on this PR?

potiuk:
H,,, I honestly think we should not invest too much in it. We are discussing about switching to Pelikan which is going to change the process completely anyway.

potiuk:
There are a number of problems with dependencies the current approach has (and it caused us some problems in the recent past) so I think this is somethign we have to do anyway soon.

Bowrna:
@potiuk Thanks for the update. I am happy to help in switching to Pelikan part :)

Bowrna:
could i close this PR @potiuk as this is not relevant if switching to Pelican?

potiuk:
yes

PR #639, opened 2022-07-16
Change protection rules for airflow-site repository

potiuk:
We pretty much never are able to review airflow-sites changes anyway
and we do not need to protect the branch as we do not release
software from it - just produce documentation.

The protection rules are now:

* you cannot push directly to `main`
* you need a PR with successful checks
* the PR must be up-to-date (i.e. based on latest main)
* no need for approval from other committers

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -32,8 +32,9 @@ github:
 
   protected_branches:
     main:
-      required_pull_request_reviews:
-        required_approving_review_count: 1
+      required_status_checks:
+        strict: true
+      required_linear_history: true
 
   ghp_branch: gh-pages
 


ephraimbuddy:
Won't this make it possible for someone to push a commit directly to main without making a pull request? If it will, I would suggest still protecting main because I have occasionally iterated changes on a repo main and tried pushing only to realize that I'm on main. Some other persons may also be facing this issue

potiuk:
> Won't this make it possible for someone to push a commit directly to main without making a pull request? If it will, I would suggest still protecting main because I have occasionally iterated changes on a repo main and tried pushing only to realize that I'm on main. Some other persons may also be facing this issue

Yep. you are absolutely right.

I changed it now, to still require a PR and status checks to pass, but no longer require an approval.

This is - I think - the best approach. We won't be able to force-push to master, we will always have to go through PR which will have to pass all the checks (I.e. the site shoudl build), and the PR will have to be based on latest `main` - keeping linear history and forcing the contributor to rebase before merge.

potiuk:
BTW. you can read more about asf.yaml at https://cwiki.apache.org/confluence/display/INFRA/Git+-+.asf.yaml+features#Git.asf.yamlfeatures-Branchprotection

ephraimbuddy (approved):

PR #641, opened 2022-07-20
Doc menu items are not collapsible #24963

anselboero:
Closes Airflow [#24963](https://github.com/apache/airflow/issues/24963[](url)) Issue.
Clicking on the the active menu's arrow will show/hide the submenus.

ezgif-4-a4b8168ba4

I noticed that the content inside the **globaltoc.html** page is repeated 2 times in the built web pages.

The first time the HTML code is inserted inside the "**content-drawer-wrapper**" div, used to render the sidebar inside the "Content" hamburger button, which is shown only when the horizontal screen reaches a small width.

Screenshot from 2022-07-20 16-31-53

The HTML code of the **globaltoc.html** file is then used to render the sidebar which is displayed under normal conditions.

This is why I included the Javascript code in a new file, I wanted to avoid it being run twice.

I'm not a front-end expert, but I don't think repeating the same Html and Css code twice to make the page responsive is considered a good practice. If you agree I could study the Sphinx documentation better and try to refactor the **sphinx_airflow_theme**, trying to remove the redundant parts.

Thank you very much for your help.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -139,6 +139,7 @@ repos:
         entry: ./site.sh lint-js
         language: system
         files: \.js$
+        exclude: ^sphinx_airflow_theme/
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v2.3.0
     hooks:
--- a/sphinx_airflow_theme/sphinx_airflow_theme/__init__.py
+++ b/sphinx_airflow_theme/sphinx_airflow_theme/__init__.py
@@ -37,5 +37,6 @@ def setup_my_func(app, config):
 def setup(app: Sphinx):
     app.add_html_theme('sphinx_airflow_theme', path.abspath(path.dirname(__file__)))
     app.add_css_file('_gen/css/main-custom.min.css')
+    app.add_js_file('js/globaltoc.js')
     app.connect("config-inited", setup_my_func)
     return {"version": "__version__", "parallel_read_safe": True, "parallel_write_safe": True}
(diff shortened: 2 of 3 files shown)


potiuk (approved):

potiuk:
WHOA !

potiuk:
Unfortunately lint is failing - you need to fix it (pre-commit will do it automatically for you so I recommend to install it)

anselboero:
Thanks a lot Potiuk for your feedback.
Unfortunately there are two points that should be considered before the pull request can be approved (They have global relevance and are not specific to this PR, if necessary I can open a new discussion).

- Error: **ESLint couldn't find a configuration file.**
I got this error as the new Js file is located in the **./sphinx_airflow_theme/sphinx_airflow_theme/static/js** folder but the configuration file used by Lint is located in the **./landing-pages** folder, which isn't an ancestor folder for the new Js file: for this reason Lint cannot check the new Js file. 
I tried to add the configuration file in an ancestor folder but this approach isn't sufficient as it's also necessary to add in the configuration file's folder the **node_modules** directory, that is at the moment created and saved by yarn in the landing-pages folder (Otherwise Lint cannot find some dependencies such as babel-eslint)

    Therefore **at the moment I think it is not possible to run Lint on Js files saved in folders other than ./landing-pages/*.**

- **Add Jquery to the Lint-Js configuration file**
Would it be possible to include the Jquery library among the environments accepted by Lint or did we decide not to use Jquery to create the site? (Although sphinx uses it)

Thank you very much and sorry if the comment may seem too long or unclear

potiuk:
You can exclude your file from being linted in the appropriate entrry of .pre-commit.yam.

potiuk:
Woohoo!

PR #645, opened 2022-07-28
Update link to lead the user directly to the documentation

potiuk:
The Python Client link led to the whole repo, where with this
change it will be directly to the README.md which is
the documentation of the client (i.e. what the user was looking for)


potiuk:
@msumit ? Does it look right :) ?

Issue #685, opened 2022-11-23
[docker-compose] v2 (deprecated v1)


seunggabi:
- https://www.docker.com/blog/announcing-compose-v2-general-availability/

PR #686, opened 2022-11-23
(#685) feat: docker compose v2

seunggabi:
- #685 
- ide fix space & eof newline
- can I rollback ?

### DIFF
--- a/docs-archive/apache-airflow/2.0.1/_sources/start/docker.rst.txt
+++ b/docs-archive/apache-airflow/2.0.1/_sources/start/docker.rst.txt
@@ -76,7 +76,7 @@ On **all operating system**, you need to run database migrations and create the 
 
 .. code-block:: bash
 
-    docker-compose up airflow-init
+    docker compose up airflow-init
 
 After initialization is complete, you should see a message like below.
 
@@ -96,7 +96,7 @@ Now you can start all services:
 
 .. code-block:: bash
 
-    docker-compose up
+    docker compose up
 
 In the second terminal you can check the condition of the containers and make sure that no containers are in unhealthy condition:
 
@@ -127,7 +127,7 @@ You can also run :doc:`CLI commands </usage-cli>`, but you have to do it in one 
 
 .. code-block:: bash
 
-    docker-compose run airflow-worker airflow info
+    docker compose run airflow-worker airflow info
 
 If you have Linux or Mac OS, you can make your work easier and download a optional wrapper scripts that will allow you to run commands with a simpler command.
 
@@ -190,7 +190,7 @@ To stop and delete containers, delete volumes with database data and download im
 
 .. code-block:: bash
 
-    docker-compose down --volumes --rmi all
+    docker compose down --volumes --rmi all
 
 Notes
 =====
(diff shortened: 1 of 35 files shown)


potiuk (approved):

potiuk:
Yeah. We can fix it everywhere now. Sufficient time passed for people to start using `docker compose`. I just wonder @seunggabi  - do you know how `dockoer compose` works with other tools such as colima using docker runtime ?

potiuk:
> @potiuk [docker/compose#9598 (comment)](https://github.com/docker/compose/pull/9598#issuecomment-1167305603)
> 
> yes. As far as I know, it guarantees backward compatibility.

Including colima?

seunggabi:
@potiuk 
let me check.
I think that, v2 is okay.

but, cross check by docker forums.
https://forums.docker.com/t/docker-compose-docker-compose-v2-support-colima-guarantees-backward-compatibility/131967

please wait for answer.
thx.

seunggabi:
https://airflow.apache.org/docs/apache-airflow/stable/howto/docker-compose/index.html#initialize-the-database

actually, now docs mixed v1 & v2 haha.


스크린샷 2022-11-23 오후 9 21 28

seunggabi:
@potiuk 

### summary
```
v2 is designed to be backward compatible in theory
I understood that we couldn’t guarantee 100% because we didn’t test everything on a 1:1 basis.
```
https://forums.docker.com/t/docker-compose-docker-compose-v2-support-colima-guarantees-backward-compatibility/131967/2

### docker forums leader said..
```
I am not sure what you need an answer to. If you want more people to read your question and try to answer, it is better to descibe the issue directly in the post instead of linkin an other forum so we can read everything, find out what the problem is, which was the question and then answer here…

Despite that I have read the issue on GitHub and I am still not sure what the question is. Is it that you would like to know if Docker Compose v1 is compatible with v2 so you can start to use docker compose command in airflow instead of docker-compose.

In theory, yes. On the other hand Docker Compose v2 had a bug from the beginning, because parsing of environment variables in the compose file had changed so previously working compose files would not always work with Compose v2. I submitted a pull request to fix it and it was eventually accepted, but I haven’t tried if that change was implemented in a Docker Compose release.

The othe rproblem could be that Docker Compose v1 supported different compose yaml versions, but there is no version attribute in a Compose V2 file so some features may work differently if the application relied on something that was or wasn’t in a specific compose yaml version…

This is just an example so I think in most of the cases you would not have problem with Compose v2, but it is not impossible. I wouldn’t rely on an intended compatibility in any application so I wouldn’t do that with Docker Compose either. If you want to make sure that everything works, you need to test every command.

The library in Docker Compose v2 didn’t have a test for multiple variables in one string and this is why the developers didn’t notice this issue,
```

image

rimelek:
I added some new notes to be more precise on the Docker forum in the topic that @seunggabi linked. So most of the compose files would work with Docker Compose v2, but not all of them. You can use the `--compatibility` flag, but I haven't tested how it works. Although all of my old compose files worked even without compatibility mode. Don't forget to check the docs about the compatibility of v1 and v2:

https://docs.docker.com/compose/cli-command-compatibility/

potiuk:
Let's see :)

PR #715, opened 2022-12-20
Adding correct reference to globaltoc.js

### LINKED ISSUE #713: Docs site making request to globaltoc.js but it doesn't exist
### What do you see as an issue?

Most pages on the airflow docs are making API calls to `globaltoc.js` and the file doesn't exist anymore leading to these requests being repetitive and exiting with 404 status code. This also results in an ugly browser console flooded with these 404 errors. 
For example:
image

### Solving the problem

The problem can be solved by disabling the calls to this endpoint so that the calls don't go and the error doesn't come

### Anything else

Issue happens at intervals when on **any** airflow doc page

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/apache/airflow/blob/main/CODE_OF_CONDUCT.md)

amoghrajesh:
Fixes #713

### DIFF
--- a/docs-archive/apache-airflow/2.5.0/_static/globaltoc.js
+++ b/docs-archive/apache-airflow/2.5.0/_static/globaltoc.js
@@ -1,0 +1,25 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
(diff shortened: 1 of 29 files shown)


amoghrajesh:
@potiuk need help on reviews here

amoghrajesh:
Following up for some review on this @potiuk

potiuk (approved):

Issue #720, opened 2023-01-05
Ecosystem Link in Nav goes to the wrong page.


johnstonmatt:
The blue "Ecosystem" button in the nav on the landing page links to:

https://airflow.apache.org/docs/apache-airflow/stable/start.html

instead of

https://airflow.apache.org/ecosystem/

This is the button:

<img width="1728" alt="Screenshot 2023-01-05 at 6 37 55 PM" src="https://user-images.githubusercontent.com/5455542/210893789-be65cd98-2ecd-4100-82ad-7a64f0ce94de.png">

potiuk:
Should be fixed in a few minutes when the sites rebuild itself.s Thanks @johnstonmatt for reporting it!

potiuk:
It's fixed now.

PR #721, opened 2023-01-06
Fix "Call to action" button to show "Install"

potiuk:
The change #678 removed unused "Install" page which had been indexed in Google but not used anywhere. However the presence of that install page was accounted for in the "landing" page where it was skipped by the code that prepared navigation bar.

Instead the "Ecosystem" was skipped rather than Install and the "Install" button was named "Ecosystem".

This change brings it back to normal by expecting the exact number of pages and by rendering the install link unconditionally

Fixes: #720


potiuk:
Looks good. Merging.

PR #740, opened 2023-02-23
Remove old versions of Airflow to save space during builds

potiuk:


Taragolis:
😮 
image

eladkal:
What versions are we removing? Such solution might cause problems for users who run on older versions and wont be able to find relevant docs

Taragolis:
I don't have any better idea.

Just one question Is it affected docs stored in https://airflow.readthedocs.io ?
image

potiuk:
> I don't have any better idea.
> 
> Just one question Is it affected docs stored in https://airflow.readthedocs.io ? image

I think not. Those are independent. And yes, lilkely redirecting the 1.10 * airflow to old versions in read the docs if we find they will remain is a good idea as temporary measure (this problem is not going away by removing 1.10* versions, it just gets delayed).

potiuk:
> What versions are we removing? Such solution might cause problems for users who run on older versions and wont be able to find relevant docs

We are removing `< 1.10.15`. If there are better ideas how to solve current generation, I would prefer it too, but I have absolutely no other idea (and absolutely no time to do it differently).

Taragolis (approved):
I have no objections as well as better idea, in the end we always have a button revert commit (if it would work in this case).

But also also would be nice to hear others

eladkal (approved):
I guess if we ever do find a better aolution we can revert the PR.. for now we must take action so newer docs will be available

potiuk:
Yeah. I think for now It's the only "stop-gap" we can implement. I hope we will fix it permanantly soon as well, otherwise we will have to start remove other archives as well.

PR #744, opened 2023-02-25
Fix wrapping of unsubscribe lists in the community page

potiuk:
The list names were badly wrapping - suggesting that the unsubscribe
address was "unsubscribe@airflow.apache.org" rather than
"users-unsubscribe"

### DIFF
--- a/landing-pages/site/content/en/community/_index.html
+++ b/landing-pages/site/content/en/community/_index.html
@@ -16,10 +16,18 @@ menu:
             </p>
             <ul class="ticks-blue mx-auto">
                 <li>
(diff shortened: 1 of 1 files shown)


potiuk:
Before:

Screenshot 2023-02-25 at 09 21 44


After:

Screenshot 2023-02-25 at 09 21 08

pierrejeambrun (approved):

pierrejeambrun:
@potiuk We might want to add a css class to the new 'ul' items so it removes the default 'grey dot' for list items.

image

pierrejeambrun:
Also it was already the case, but the long emails are breaking the layout for mobile, and this is making it worse:
image

Adding an `overflow-wrap: anywhere` property to the list item with emails should make it break correctly on small screens:
image

potiuk:
I am not that good of in css and javascript to know where and how to do it :)

potiuk:
But yeah. it can be done better for sure :D

pierrejeambrun:
No worries I'll take a look thank you :)

PR #761, opened 2023-04-03
Update robots.txt to crawl only latest version of documentation

vincbeck:
See issue [#29283](https://github.com/apache/airflow/issues/29283).

The current robots.txt allows all pages to be indexed by search engines. As a a consequence, some old version of documentation ends up being indexed which can be confusing for users. They end up reading old version of documentation thinking it is the latest one. An example of user getting confused [here](https://apache-airflow.slack.com/archives/CSS36QQS1/p1680524708746749).

Which this change, all pages under `/docs/apache-airflow/` but `/docs/apache-airflow/stable/` **will stop being crawled by search engines**. !! Please confirm this is fine since this change is quite impactful !!

### DIFF
--- a/landing-pages/site/layouts/robots.txt
+++ b/landing-pages/site/layouts/robots.txt
@@ -1,0 +1,5 @@
+User-agent: *
+# Block all bots from accessing the documentation but the stable version
+# We do not want the search engines to index the documentation other than the lastest version
+Disallow: /docs/apache-airflow/
+Allow: /docs/apache-airflow/stable/


o-nikolas:
Is there a way to provide a weight to the stable docs and still keep the older docs indexed (but they will just be less likely to show up unless searching for terms only contained in older docs)? Or is it only enable/disable?

vincbeck:
As far as I know it is only enable/disable. See documentation [here](https://developers.google.com/search/docs/crawling-indexing/robots/create-robots-txt). And my personal opinion is I rather not index old docs so users are always redirected to the most up to date docs and if they want to read documentation about a specific version they can select it from the left dropdown

o-nikolas:
> if they want to read documentation about a specific version they can select it from the left dropdown

Yeah, for sure, they'll still be able to find the docs, but if the user is knowingly running an older version of Airflow (and with this change older can mean just a single release back)it is helpful to be able to google search for terms that may only exist in the docs for the version which apply to you.

But if it's not possible then I suppose it's moot. I'm interested to hear what others think, because I agree with you that it's quite an impactful change! Perhaps it's worth an email on the dev list?

potiuk:
I think I'd go half-way. Remove All 2.0, 2.1, 2.2, 1.10 docs from the index. That would likely be a much better choice. We already have somewhat soft deprecation mechanism via providers. It's not "you should upgrade" but "if you don't, latest provider releases will not be installable for you". In May we bump it to 2.4, so we can add 2.3 to the exclusions too. 

While this is not a "strict" deprecation mechanism, I think it's as good as can be to push people in the direction of upgrade. I think not indexing those docs is a good reflection of our more or less gentle suggestion to upgrade and reflecting that in the index is a good idea.

vincbeck:
> > I think I'd go half-way. Remove All 2.0, 2.1, 2.2, 1.10 docs from the index. That would likely be a much better choice. We already have somewhat soft deprecation mechanism via providers. It's not "you should upgrade" but "if you don't, latest provider releases will not be installable for you". In May we bump it to 2.4, so we can add 2.3 to the exclusions too.
> 
> Yeah I am happy with that. But we should add the big red banner with warning that this is an old version and provide link for stable version.
> 
> Example with SQLAlchemy: (https://docs.sqlalchemy.org/en/14/intro.html#installation)
> 
> image
> 
> Example with [Astro SDK](https://astro-sdk-python.readthedocs.io/en/1.5.2/) (that uses ReadTheDocs): image

Agree with this. Though, I was thinking splitting the change in two PRs, this one exclusively about the robots.txt change and then I (or someone else?) will create another one to add the banner

potiuk:
> Yeah I am happy with that. But we should add the big red banner with warning that this is an old version and provide link for stable version.

Yeah would be nice if someone is brave enough to dive into the depths of our docs building system to do such conditional banner.

vincbeck:
I found a way to fetch all versions from providers and Airflow. I am using the file `packages-metadata.json`. That way we can apply some deterministic rules to decide whether a given version should be indexed. The current implementation disallows all versions but the last one. **I understand this is not what we want and just did that for example**. Here is the
[robots.txt](https://github.com/apache/airflow-site/files/11151148/robots.txt) file generated with this implementation.

I would like to use this PR to discuss and decide on which condition a given version (for Airflow and providers) should be indexed by search engines.
- Airflow: @potiuk suggested to have an upper limit (currently `2.3` and will potentially bump up to `2.4` in May). Everything **below** this upper limit would NOT be indexed by search engines. This upper limit is the minimum Airflow version needed to install latest provider releases
- Providers. Hard question. My personal opinion is we should come up with a rule which applies to all providers. Having provider specific rules would be too complex/time spending to manage. That being said I dont have strong opinion on which rule we should have for providers. e.g. Only versions which belong the latest major version? Only versions which belong the 2 latest major versions? WDYT?

vincbeck:
Interesting. I did know that. It looks like a very good fit for our problem indeed. The doc actually mention "Don't use the robots.txt file for canonicalization purposes.". And by reading the doc and definition of canonicalization, it looks like our problem here is very canonicalization. Any other opinion?

potiuk:
That's an interesting idea @mik-laj :) . It's not exactly why canonical was created for, but it might do it in our case. However this is significantly more complex to implement because not only we would have to post-process historical pages and add "canonical" link, but we would have to continuously update old pages as they get older - note that "stable" is a symbolic link to the last version, so in effect the stable links would refer to themselves,

This on it's own is not a bad practice - it seems to be even recommended https://www.searchenginejournal.com/google-self-referencing-canonicals-are-not-critical/312619/  but then we either post-process the docs when some new version is released or we are limiting ourselves strictly to only indexing the very last patchleve version for airflow or providers.

IMHO - If we choose to only index the last patchlevel - then using canonical makes sense. If we choose another scheme, this is a lot of hassle.

vincbeck:
Basically the canonical link for each package would be the stable one right?

Example:
- https://airflow.apache.org/docs/apache-airflow/stable/index.html is the canonical link for Airflow documentation. https://airflow.apache.org/docs/apache-airflow/2.5.3/index.html is not the canonical link even though, today, it goes to the same page

Is it was you are suggesting?

I do not know much about documentation generation in Airflow and the work required in order to do so but it seems doable. For each package, we just need to add a `<link tel="canonical"` which specifies the stable link as the canonical one. Am I wrong?

vincbeck:
Alright, I'll do that then

vincbeck:
I'll use this PR to add the canonical link to newly generated docs. Of course I'll revert the changes I mades to robots.txt as well. I'll create another PR to update archived docs

vincbeck:
I updated the PR to include the canonical link in newly generated doc. HOWEVER, I could not find any way to really test the doc generation with this new layout. The only way I could find was using `demo` but I'd like to generate the actual docs with this new layout. Anyone knows how can I do that?

potiuk:
This was always a bit of mistery to me especially when you make change to a theme. But:

* Building and installing theme package locally https://github.com/apache/airflow-site/tree/main/sphinx_airflow_theme
* Instruction to build the docs https://github.com/apache/airflow-site/blob/main/CONTRIBUTE.md#instruction 

I think some of the docs/versions are a bit outdated and might require some trickery to install on newer systems though

potiuk:
> Followed 1. and 2. only. The script docs/build.sh does not exist anymore in Airflow repository so I ran these commands instead to build the doc and visualize them:

Look what breeze does in the `developer_commands.py` for build_docs:
```
    cmd = [
        "docker",
        "run",
        "-t",
        *extra_docker_flags,
        "--pull",
        "never",
        ci_image_name,
        "/opt/airflow/scripts/in_container/run_docs_build.sh",
        *doc_builder.args_doc_builder,
```

or you can use `--dry-run` when running breeze command and it will show you exactly what it does.

You can enter breeze and run the commnd manually, and install the sphinx theme locally before I believe (possibly needs some modifications of the shell script)

potiuk:
BTW. @vincbeck  I think it should be **just** enough to copy the theme to /files folder, enter install the theme from  there `pip install /files/your.whl` and then run the " /opt/airflow/scripts/in_container/run_docs_build.sh" script with the right args .... You should not even need to run the `--dry-run`  or copy&paste anything.

The `run_docs_build.sh` has `--help` and it will print you the options (they are basically the same as in breeze command). 

So I think the issu you had was basically that the original script is moved :)

potiuk:
> So I think the issue you had was basically that the original script is moved :)

I hope the result of this exercise will also be updating the docs :D

vincbeck:
@potiuk I am not able to locate this `.whl` file. Any hint on where I can find it?

potiuk:
> @potiuk I am not able to locate this `.whl` file. Any hint on where I can find it?

You should build it - I think one of the steps (with site.sh script does it)  https://github.com/apache/airflow-site/blob/main/sphinx_airflow_theme/README.md (and if not that's I guess pretty standard python package that can be built with `python setup.py` ?

vincbeck:
I could finally tested my changes. It was correct. I updated the doc and added a new section to explain the steps I followed so if anyone needs to do the same thing, it should be easy

potiuk:
Coool! One more thing - can you please update version of the "sphinx-theme" package @vincbeck - then I **think** I should be able to publish it to PyPI (I think I have permissions :D).

vincbeck:
Done!

potiuk (approved):

potiuk:
Uploaded : https://pypi.org/project/sphinx-airflow-theme/0.0.12/ , Merging.

potiuk:
I **hope** it's going to work.

PR #794, opened 2023-06-08
Building framework to allow multiple types of back referencing docs

amoghrajesh:
The script needs to be more robust to accept an argument to be able to generate back referencing docs for airflow, helm, and providers.

This PR tries to extend the initial code to build a framework that can accept such parameters and conditionally do such tasks

### DIFF
--- a/post-docs/README.md
+++ b/post-docs/README.md
@@ -1,0 +1,9 @@
+# Post Documentation Process
+
+In order to support backward compatibility of the documentation generation process, this additional step is needed to add refresh HTMLs for older released docs.
+The issue persists across helm, provider, and regular airflow docs.
+
+To generate these back referencing(refresh HTMLs), run the script in the following manner:
+```commandline
+python add-back-reference.py [airflow | providers | helm]
+```
--- a/post-docs/README.md
+++ b/post-docs/README.md
@@ -3,7 +3,7 @@
 In order to support backward compatibility of the documentation generation process, this additional step is needed to add refresh HTMLs for older released docs.
 The issue persists across helm, provider, and regular airflow docs.
 
-To generate these back referencing(refresh HTMLs), run the script in the following manner:
+To generate these back referencing (refresh HTMLs), run the script in the following manner:
 ```commandline
 python add-back-reference.py [airflow | providers | helm]
 ```
--- a/post-docs/add-back-references.py
+++ b/post-docs/add-back-references.py
@@ -18,6 +18,7 @@ import enum
 import logging
 import os
 import sys
+from pathlib import Path
 from urllib.request import urlopen
 import semver
 
@@ -29,12 +30,6 @@ docs_archive_path = "../docs-archive"
 airflow_docs_path = docs_archive_path + "/apache-airflow"
 helm_docs_path = docs_archive_path + "/helm-chart"
 providers_docs_path = docs_archive_path + "/apache-airflow-providers"
-
-# version where changes were introduced in docs structure
-new_airflow_docs_version = "2.5.1"
-new_helm_docs_version = "1.6.0"
-# change this correctly
-new_providers_docs_version = "0.0.0"
 
 
 # types of generations supported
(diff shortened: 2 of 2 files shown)


post-docs/README.md
```diff
@@ -0,0 +1,9 @@
+# Post Documentation Process
+
+In order to support backward compatibility of the documentation generation process, this additional step is needed to add refresh HTMLs for older released docs.
+The issue persists across helm, provider, and regular airflow docs.
+
+To generate these back referencing(refresh HTMLs), run the script in the following manner:
```
vincbeck, line 6:
```suggestion
To generate these back referencing (refresh HTMLs), run the script in the following manner:
```

post-docs/add-back-references.py
```diff
@@ -76,30 +88,49 @@ def create_back_reference_html(back_ref_url, path):
...
+    logging.Logger.error("missing required arguments, syntax: python add-back-references.py [airflow | providers | "
+                         "helm]")
+
+type_generation = sys.argv[1]
+if type_generation == airflow_type:
+    generate_back_references(airflow_redirects_link, airflow_docs_path, new_airflow_docs_version)
+elif type_generation == helm_type:
+    generate_back_references(helm_redirects_link, helm_docs_path, new_helm_docs_version)
+elif type_generation == provider_type:
+    pass
```
vincbeck, line 133: Is it work in progress?

pierrejeambrun: Maybe for providers this does not apply, a log explaining that could help ?

post-docs/README.md
amoghrajesh: Sure, adding this change

post-docs/add-back-references.py
```diff
@@ -14,16 +14,28 @@
...
 import os
+import sys
 from urllib.request import urlopen
 import semver
 
-docs_link = "https://raw.githubusercontent.com/apache/airflow/main/docs/apache-airflow/redirects.txt"
+# types of generations supported
+airflow_type = "airflow"
+helm_type = "helm"
+provider_type = "providers"
```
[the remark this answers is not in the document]
amoghrajesh: Yes, good idea. Let me make that change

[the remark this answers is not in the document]
vincbeck: You can remove then now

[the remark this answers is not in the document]
amoghrajesh: Oops, missed that. Removed it now

vincbeck (approved):

post-docs/add-back-references.py
```diff
@@ -14,16 +14,34 @@
...
 docs_archive_path = "../docs-archive"
-apache_airflow_path = docs_archive_path + "/apache-airflow"
-stable_version_path = apache_airflow_path + "/stable.txt"
-new_docs_version = "2.5.1"
+airflow_docs_path = docs_archive_path + "/apache-airflow"
+helm_docs_path = docs_archive_path + "/helm-chart"
+providers_docs_path = docs_archive_path + "/apache-airflow-providers"
+
+# version where changes were introduced in docs structure
+new_airflow_docs_version = "2.5.1"
```
potiuk, line 34: Question: Can we do it regardless for latest versions ?  I believe (correct me if I am wrong please) that when you get the "latest" version, and there are no redirects we could check if the file that we want to create redirect from exists already and skip if it does - this way we will not override "target" files, but we will be able to re-run it at any time from latest version, making it effectively idempotent.

Smth like: 

```

def create_back_reference_html(back_ref_url, path):
    content = get_redirect_content(back_ref_url)
   if Path(path).exists(): # <-- check if the target file exists already
      return 
    # Creating an HTML file
    with open(path, "w") as f:
        f.write(content)
```

amoghrajesh: Very nice point, let me try out this approach.

amoghrajesh: Fetching and filtering by version seems like an additional overhead which we should try to avoid as much as possible.

amoghrajesh:
@potiuk Can i get a review on this when you have some time? We can start working on the next step once we merge this one in

potiuk (approved):

potiuk:
LGTM

PR #799, opened 2023-06-12
Extending the back reference script to support providers

amoghrajesh:
With this change, this script can be used to generate back reference docs even for airflow providers

### DIFF
--- a/post-docs/add-back-references.py
+++ b/post-docs/add-back-references.py
@@ -18,6 +18,7 @@ import enum
 import logging
 import os
 import sys
+import tempfile
 from pathlib import Path
 from urllib.error import URLError
 from urllib.request import urlopen
@@ -41,14 +42,18 @@ class GenerationType(enum.Enum):
 
 
 def download_file(url):
-    try:
-        filedata = urlopen(url)
-        datatowrite = filedata.read()
-
-        with open('redirects.txt', 'wb') as f:
-            f.write(datatowrite)
-    except URLError as e:
-        log.warning(e)
+    with tempfile.TemporaryDirectory() as tmpdir:
+        temp_dir = Path(tmpdir)
+        file_name = temp_dir / "redirects.txt"
+        try:
+            filedata = urlopen(url)
+            data = filedata.read()
+            with open(file_name, 'wb') as f:
+                f.write(data)
+            return True
+        except URLError as e:
+            log.warning(e)
+            return False
 
 
 def construct_mapping():
@@ -106,7 +111,10 @@ def create_back_reference_html(back_ref_url, path):
 
 
 def generate_back_references(link, base_path):
-    download_file(link)
+    is_downloaded = download_file(link)
+    if not is_downloaded:
+        log.warning('skipping generating back references')
+        return
     old_to_new = construct_mapping()
 
     versions = [f.path.split("/")[-1] for f in os.scandir(base_path) if f.is_dir()]
--- a/post-docs/add-back-references.py
+++ b/post-docs/add-back-references.py
@@ -22,14 +22,30 @@ from pathlib import Path
 from urllib.request import urlopen
 import semver
 
+log = logging.getLogger(__name__)
+
 airflow_redirects_link = "https://raw.githubusercontent.com/apache/airflow/main/docs/apache-airflow/redirects.txt"
 helm_redirects_link = "https://raw.githubusercontent.com/apache/airflow/main/docs/helm-chart/redirects.txt"
-providers_redirect_link = "populate-this"
 
 docs_archive_path = "../docs-archive"
 airflow_docs_path = docs_archive_path + "/apache-airflow"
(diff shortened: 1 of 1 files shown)


post-docs/add-back-references.py
```diff
@@ -22,14 +22,30 @@
...
 airflow_redirects_link = "https://raw.githubusercontent.com/apache/airflow/main/docs/apache-airflow/redirects.txt"
 helm_redirects_link = "https://raw.githubusercontent.com/apache/airflow/main/docs/helm-chart/redirects.txt"
-providers_redirect_link = "populate-this"
 
 docs_archive_path = "../docs-archive"
 airflow_docs_path = docs_archive_path + "/apache-airflow"
 helm_docs_path = docs_archive_path + "/helm-chart"
-providers_docs_path = docs_archive_path + "/apache-airflow-providers"
+
+# if any new provider is added with redirects, add it to this list
```
potiuk, line 34: Since we are downloading the redirects from GitHub already can we simply loop over all directories in "docs-archive" and react to 404 not found by skipping a provider? that should be pretty fast.

potiuk: Why: the less chance we have to forget to update something, the better. This is for example why in my in-progress PR about sbom generation, I initially had a hard-coded list of Airflow versions - but eventually I settled on retriving them from tags in the git remote "apache/airflow" repo:
 https://github.com/apache/airflow/pull/31835/files#diff-54ea68b5e1e61ac0437df0ab7d99dc508595323b11847a957931e0d9007962d7R1241

amoghrajesh: That was my initial idea too. But how do I get the list of directories in GitHub through my script?

potiuk (commented):
Just one small comment about automation of list of providers :). Looks cool

post-docs/add-back-references.py
potiuk: No need. Just list those in `docs-archive`

amoghrajesh: Good idea. Missed that.
Yeah I agree with you on this, pushing a fix for that :)

post-docs/add-back-references.py
```diff
@@ -40,11 +41,14 @@ class GenerationType(enum.Enum):
...
+    try:
+        filedata = urlopen(url)
+        datatowrite = filedata.read()
 
-    with open('redirects.txt', 'wb') as f:
-        f.write(datatowrite)
+        with open('redirects.txt', 'wb') as f:
+            f.write(datatowrite)
+    except URLError as e:
+        log.warning(e)
```
potiuk, line 51: Should not we also skip generation in this case?

potiuk: (I think what is going to happen with this version is that will keep on using the redirects.txt retrieved from previous successfully downloaded redirect.txt ? I guess it would be great to  do two things:

1) make sure that the redirects.txt is named differently depending on a provider  (just in case of programming mistakes) 

2) store them somewhere where they will be deleted (tmp dir)

3) if download fails, skip the whole `generate`

amoghrajesh: Very good point. Will handle these points soon and re request for review :)

amoghrajesh:
@potiuk handled the following issues in the latest commit:
- Moved the redirects.txt to a temp directory, cleaning won't be an issue
- Skipping generation in case fetching of the file fails, ie: file doesn't exist, or if redirects are not applicable to that provider

This should handle most edge cases imo. Can you have a round of review when free?

post-docs/add-back-references.py
```diff
@@ -40,11 +42,18 @@ class GenerationType(enum.Enum):
...
 
 def download_file(url):
-    filedata = urlopen(url)
-    datatowrite = filedata.read()
-
-    with open('redirects.txt', 'wb') as f:
-        f.write(datatowrite)
+    with tempfile.TemporaryDirectory() as tmpdir:
+        temp_dir = Path(tmpdir)
+        file_name = temp_dir / "redirects.txt"
```
potiuk, line 47: The downloaded redirects.txt file will not be used - it is notpassed anywhere. I see that in case of construct_mapping method it still opens "current working dir" redirects.txt - which is not going to work - if you downliad the file here, it should be passed on to processing

amoghrajesh: Thanks for the feedback. I realised this fault while testing and I have fixed it in the latest commit

potiuk: LGTM now!

potiuk:
I guess time to incorporate it in provider's release docs @eladkal  @amoghrajesh :)

amoghrajesh:
Nice. How do we do that?

potiuk:
Update steps t follow in in https://github.com/apache/airflow/blob/main/dev/README_RELEASE_PROVIDER_PACKAGES.md

PR #820, opened 2023-06-25
Add (TM) in airflow documentation for latest release

potiuk:


potiuk:
I cherry-picked https://github.com/apache/airflow/pull/31977 to v2-6-test and used it to regenerate the docs so that we can continue with registration of the Airflow trademark,

PR #822, opened 2023-06-27
Announce helm chart 1.10.0

jedcunningham:


potiuk (approved):
Checked the links - all look good.

BTW. For those selected few who still remember, the "1.10.0" version has some very, very, very strong negative connections in Airflow world.... 

I would really love to get `1.11.0` or smth like that SOOON :rofl:

potiuk:
> BTW. For those selected few who still remember, the "1.10.0" version has some very, very, very strong negative connections in Airflow world....

BTW. It's even stronger than the upcoming "Airflow 2.7"

PR #830, opened 2023-07-09
Adds generation of the new changelog and security links

potiuk:
Add automated generation of the security and changelog links.

* Added a few renames of variables and types to make it easier to understand what's going
* Converted logs into rich's print with colours to better distinguish info, warnings, changes
* Converted dict into tuple, because we could have two new pages point to the same old page (and it was not possible when old page was key in a dictionary
* Hard-coded changelog and security links because they are new and have no redirects - that also required to process all versions of all providers, because they should be processed even without redirects.

### DIFF
--- a/post-docs/README.md
+++ b/post-docs/README.md
@@ -7,3 +7,6 @@ To generate these back referencing (refresh HTMLs), run the script in the follow
 ```commandline
 python add-back-reference.py [airflow | providers | helm]
 ```
+
+Before running the script, make sure to install requirements from `requirements.txt` file into
+your virtual environment. For example by `pip install -r requirements.txt`.
--- a/post-docs/add-back-references.py
+++ b/post-docs/add-back-references.py
@@ -14,16 +14,15 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
+from __future__ import annotations
 import enum
-import logging
 import os
 import sys
 import tempfile
 from pathlib import Path
 from urllib.error import URLError
 from urllib.request import urlopen
-
-log = logging.getLogger(__name__)
+from rich import print
 
 airflow_redirects_link = "https://raw.githubusercontent.com/apache/airflow/main/docs/apache-airflow/redirects.txt"
 helm_redirects_link = "https://raw.githubusercontent.com/apache/airflow/main/docs/helm-chart/redirects.txt"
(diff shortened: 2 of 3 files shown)


potiuk:
cc: @amoghrajesh: Few updates to the script:

Now it looks like this:


Screenshot 2023-07-09 at 20 50 06

amoghrajesh:
I'll go through this in my IST time tomorrow

post-docs/README.md
```diff
@@ -7,3 +7,6 @@ To generate these back referencing (refresh HTMLs), run the script in the follow
 ```commandline
 python add-back-reference.py [airflow | providers | helm]
 ```
+
+Before running the script, make sure to install requirements from `requirements.txt` file into
+your virtual environment.
```
amoghrajesh, line 12: Can we provide the command here?
`pip install -r requirements.txt` here?

post-docs/requirements.txt
```diff
@@ -0,0 +1 @@
+rich
```
amoghrajesh, line 1: I think we also need to add `urllib` here

post-docs/add-back-references.py
```diff
@@ -149,9 +154,8 @@ def generate_back_references(link, base_path):
...
-    for p in all_providers:
-        log.info("processing airflow provider: %s", p)
-        generate_back_references(get_github_redirects_url(p), get_provider_docs_path(p))
+    for provider in all_providers:
+        print(f"Processing airflow provider: {provider}")
+        generate_back_references(get_github_redirects_url(provider), get_provider_docs_path(provider))
 else:
-    log.error("invalid type of doc generation required. Pass one of [airflow | providers | "
-              "helm]")
+    print("[red]Invalid type of doc generation required[/]. Pass one of [airflow | providers | helm]")
```
amoghrajesh, line 161: nit: `[red]Invalid type of doc generation requested[/]. Pass one of [airflow | providers | helm]`

amoghrajesh (commented):
I have a few nits. Overall things look good!

potiuk:
Updated.

potiuk:
BTW. Eventually I will incorporate it all into a new `breeze publish-docs` command.

amoghrajesh:
Oh nice. Can I take that up if you don't mind? Want to understand more about that process.
(Maybe you can guide me)

amoghrajesh (approved):
Looks awesome. Thanks!
LGTM +1

PR #843, opened 2023-08-06
Fix Sphinx Airflow Theme to work for Sphinx 7

potiuk:
Still the theme/CSS is pretty bad but at least it can build on CI with those changes.


jscheffl:
I assume this can be closed meanwhile?

potiuk:
Yes.

PR #876, opened 2023-10-14
Add snapp use case

tadeha:
We at Snapp would like to share how using Airflow has helped us save time and resources on various tasks.


tadeha:
We at Snapp would like to share how using Airflow has helped us save time and resources on various tasks.

potiuk:
You need to fix the static checks (pre-commit might fix it for you automatically) seems there are spaces at the end of lines.

tadeha:
> You need to fix the static checks (pre-commit might fix it for you automatically) seems there are spaces at the end of lines.

I made changes to the code by using the examples and logos that were added to the repository.

PR #907, opened 2023-12-12
Fix daskexecutor provider docs

eladkal:
There is no daskexecutor 1.1.1
It seems that doc build created this package though it wasn't part of the release wave - I will investigate and fix the problem with the doc build


potiuk:
Well. We SHOULD release daskexecutor 1.1.1 as part of this wave. That's the whole point of having the last release which will have "removed" notice in the `PyPI` description - there is no way to change the description in `PyPI` without releasing the new version of the package.

This is defined as part of the removal process.

potiuk:
So both - releasing 1.1.1 package and documentation for it are absolutely intended and needed.

eladkal:
> Well. We SHOULD release daskexecutor 1.1.1 as part of this wave. That's the whole point of having the last release which will have "removed" notice in the `PyPI` description - there is no way to change the description in `PyPI` without releasing the new version of the package.
> 
> This is defined as part of the removal process.

No...
we are releasing 1.2.0 because of min airflow version of 2.6
see https://github.com/apache/airflow/pull/36190

potiuk:
Ah then 1.2.0 is wrong. We should not bump Airflow min version for daskexecutor. Ths "upgrade min airflow" version will sklp bumping it at least. 

We should not bump min-airflow version for this package - it should be identical as the last released version with the only difference being the description in `PyPI`. There is no particular reason why we should bump it for such a package.

eladkal:
> We should not bump min-airflow version for this package - it should be identical as the last released version with the only difference being the description in PyPI. There is no particular reason why we should bump it for such a package.

That means modifying pre-commit and enabling 2.5.0 back
This is an edge case of an edge case that the removal and bump min airflow version are closed to each other.

> We should not bump min-airflow version for this package 

Why do you feel strongly about it? what is the down side?

eladkal:
Following our discussion closed due to https://github.com/apache/airflow/pull/36191 
Thanks @potiuk

Issue #909, opened 2023-12-13
Conflict Between Yarn Installation Locally and Hadoop YARN


cpprhtn:
**Description:**
Encountering a conflict between the `Hadoop YARN` environment and the Yarn package manager required for building the Airflow website.

To resolve this conflict, it is recommended to use the alias `yarnpkg` when installing Yarn.

**References:**
https://github.com/yarnpkg/yarn/issues/673
https://github.com/yarnpkg/yarn/commit/cefa9a368dd3df1a1f1ecd9e50de9c5fabefbc92

pierrejeambrun:
Closing, cf https://github.com/apache/airflow-site/pull/910 for more context.

PR #910, opened 2023-12-13
Update yarn to yarnpkg to avoid conflicts with Hadoop

cpprhtn:
Fix: Resolve conflict with Hadoop YARN by using yarnpkg alias

While attempting to install the required yarn locally, a conflict arose due to the presence of Hadoop's yarn. To address this issue, the yarnpkg alias was utilized to ensure a smooth installation.

issues: #909

### DIFF
--- a/CONTRIBUTE.md
+++ b/CONTRIBUTE.md
@@ -103,7 +103,7 @@ nvm use 16
 
 To install yarn, run the following command:
 ```bash
-npm install -g yarn
+npm install -g yarnpkg
 ```
 
 To install hugo on Debian, run the following command:
@@ -132,7 +132,7 @@ nvm use 16
 
 To install Yarn, run the following command:
 ```bash
-brew install yarn
+brew install yarnpkg
 ```
 
 ### Static checks


potiuk:
Could you please add link to the page that explains why yarnpkg is good ? You mentioned that it is `recommended` to use it in the issue?

cpprhtn:
On my local page, the yarn of Hadoop was installed first, and it is a different package from the yarn for building this website.

Therefore, you had to use the command yarnpkg to install this webpage safely.

For more information, please refer to the following link

https://github.com/yarnpkg/yarn/issues/673

Thank you for your question @potiuk

pierrejeambrun:
I would say that this is specific to your installation and that the doc does not need to be updated.

It is obvious to which `yarn`  we are referencing in this guide, but we could make it more explicit if that was a source of confusion.

I think that aliasing should be user responsibility depending on its configuration.

potiuk:
> I would say that this is specific to your installation and that the doc does not need to be updated.

Agree with @pierrejeambrun

cpprhtn:
thanks to review
I agree too 👍

PR #911, opened 2023-12-13
Reorganize channels on community page

merobi-hub:
This would reorder the communication channels on the community page.

Specifically: 

- It could be argued that text on the community page does not reflect the way the community resources are currently being used and managed, so this change revises the text there to prioritize Slack and the dev list and moves the user list info to the question section on that page. 
 
What do you think about our presentation of these resources on the site? It seems like this content could be updated, but let's discuss.

Update: in light of the conversation here and on the dev list, the resources have been reorganized according to persona -- one column for dev-centric resources and one for user-centric ones.

Update Jan. 5: in response to helpful feedback from @potiuk , subheadings have been added and the resources have been distributed more equally for a more balanced layout.

Please see the screenshot for what the latest changes would look like.

screencapture-0-0-0-0-3000-community-2024-01-05-09_44_57


potiuk:
I am not sure if we want to put Slack as first and foremost in the community. While devlist is much harder to join and less popular, this is the only official and "authoritative" place where things get discussed and agreed to  - and (so far) this is the only place that the ASF projects are supposed to make decisions and where important things and announcements happen 

"If it did not happen on the devlist, it did not happen"

Slack is much more casual of course and where things are happening as well, but  I would even say GitHub discussions, issues and PRs is where much more "community" things happen. It's much more akin to the "devlist" - where smaller decisions are made in PRs and where issues and discussions about potential future features or current problems happen.

The problem with Slack is that it is ephemeral and with much more "synchronous" expectations and pressures (i.e. discussions basicaly disappear when not answered synchronously  - no-one will generall look back at the past discussions and questions after they move out behind the edge of your screen).  

Both devlist and github are mostly asynchronous in nature - and asynchronous communication is very important in ASF communities.

I am not sure putting Slack as first thing there is a good idea because of that. I am not sure if I have a very constructive approach how to do it differently.

Maybe an idea we should split this tab into two - signalling that we have things which is mostly for users seeking for help and troubleshooting (with slack first and foremost followed by github issues and discussions and ending at users@airfflow.apache.org) and another tab where we have a different set of links for "creators" - i.e. devlist, github PRs, github issues, github discussions. There is significant overlap of both of course, but maybe we should recognise the fact that "users" should have a different "hierarchy of needs" here than "creators" ?

Even the ASF standard "dev@" vs. "users@" follows that pattern.

Speaking of which @merobi-hub -> I think this one is a perfect candidate to start a devlist discussion about. I think, if you want this PR to proceed and get the right approach, the idea of changing navigation on our website is definitely not a decision any single or even few maintainers should take in a PR, this should be a discussion on devlist and I think you should start one and explain the motivations behind the change and what you think we are going to get by that.

merobi-hub:
Thank you for your comment and perspective @potiuk . Personally, a two-column arrangement of resources for "creators" and "users" with the dev list and Slack side-by-side makes a lot of sense to me, but I agree that expanding the discussion to the dev list would be worthwhile. I'll include the idea among others in my post there.

merobi-hub:
@potiuk I wonder what you think of this new version. As far as headings and labels, I think less is more but perhaps headings would help here. I don't see a way around the unequal column heights, unfortunately, but maybe there is one.

potiuk:
Screenshot(s) :) ?

merobi-hub:
@potiuk the screenshot in the first comment has been updated to reflect the latest changes

potiuk:
Oups :)

potiuk:
I like it a lot. How about a few changes:

1. adding headers like "Want to contribute?" and "Are you a user? 
2. Move the "Start a discussion" and "Report a bug" to under "Are you a user?" 
3. Rename "Add new feature" to "Propose new feature" and move it to under "Are you a user?"
4. Remeve "fix a bug" (it's the same as "Report a bug" really) - maybe merge the two - the last paragraphf rom "Fix a bug" about encouraging to propose a fix should be added to "Report a bug".
5. 

That would balance it a bit and make clearer that we have two personas we are addressing it to (and who the personas are).

merobi-hub:
Thanks @potiuk . I agree with you that adding headings and distributing the resources more equally will improve the layout. Implemented the changes -- see the screenshot for how they would look.

potiuk:
Looks great!

potiuk (approved):

potiuk:
ok. Let me merge it - I think no harm done... And we can always improve :)

PR #933, opened 2024-01-12
Update _index.md

dabla:
Added new entry in the Async Providers for the Airflow MS Graph API provider

### DIFF
--- a/landing-pages/site/content/en/ecosystem/_index.md
+++ b/landing-pages/site/content/en/ecosystem/_index.md
@@ -97,6 +97,8 @@ If you would you like to be included on this page, please reach out to the [Apac
 
 [Airflow Kafka Provider](https://github.com/astronomer/airflow-provider-kafka) - Apache Airflow Kafka provider containing Deferrable Operators & Sensors.
 
+[Airflow MS Graph API Provider](https://pypi.org/project/apache-airflow-providers-msgraph) - A collection of Async Operators and Triggers for the MS Graph API using the official [msgraph-sdk-python](https://github.com/microsoftgraph/msgraph-sdk-python) client.
+
 &nbsp;
 
 ## Third Party Airflow Helm Charts


Taragolis (approved):

o-nikolas (approved):

potiuk:
Well. one comment here @dabla - I think you should not publish it under `apache-airlfow-providers`. While we do not have a way to prevent it - this might be misleading to publish it with the `apache-airflow-` prefix, that it is an Apache Software Foundation project. 

Since you are managing it on your own - I would strongly suggest you to choose a different name and remove the `apache-airflow-*` or - even better - remove it and pass the ownership of the project to  the Apache Airflow organization so that it is not misused. 

This might change in the future, where organization and provenance of the packages in PyPi is more visible, also we do not have written rules about it, so this is more of a kind request rather than requirement, but I think it would be good to follow this approach - that is currently  pretty much universal - where `apache-airflow-` prefix is for the projects that are managed by the Apache Airflow PMC.


Would that be possible @dabla ?

potiuk (changes requested):
Marking that request changes until we hear back from @dabla

dabla:
> Well. one comment here @dabla - I think you should not publish it under `apache-airflow-providers-*` name. While we do not have a way to prevent it - this might be misleading to publish it with the `apache-airflow-` prefix, that it is an Apache Software Foundation project.
> 
> Since you are managing it on your own - I would strongly suggest you to choose a different name and remove the `apache-airflow-*` or - even better - remove it and pass the ownership of the project to the Apache Airflow organization so that it is not misused.
> 
> This might change in the future, where organization and provenance of the packages in PyPi is more visible, also we do not have written rules about it, so this is more of a kind request rather than requirement, but I think it would be good to follow this approach - that is currently pretty much universal - where `apache-airflow-` prefix is for the projects that are managed by the Apache Airflow PMC.
> 
> Would that be possible @dabla ?

Hello Jarek (@potiuk),

Well if possible I would like to hand over the provider to the Apache Airflow project, as this was the whole point from the start, at least if there is interest in it...  otherwise I will rename the provider and remove the apache prefixed one from pypi.  I started a discussion about that operator at the end of last year at the Airflow project: https://github.com/apache/airflow/discussions/36315

While awaiting a decission about that discussion (a vote is required if I understood well), I opted to already deploy it to pypi so it is open source and others can profit from it as well and maybe even improve it and do contributions.  If Airflow would take over the provider that would of course be even better.

We have multiple DAG's interacting with MS Graph API endpoints, and as we want to avoid custom code as much as possible as we have to handle lot's of data due to paging.  We though of implementing an operator for it using the official python client from [Microsoft](https://github.com/microsoftgraph/msgraph-sdk-python), that way we can simplify our dags and remove custom code as much as possible.

potiuk:
Please start discussion about it at the devlist https://airflow.apache.org/community/

dabla:
> We have multiple DAG's interacting with MS Graph API endpoints, and as we want to avoid custom code as much as possible as we have to handle lot's of data due to paging. We though of implementing an operator for it using the official python client from [Microsoft](https://github.com/microsoftgraph/msgraph-sdk-python), that way we can simplify our dags and remove custom code as much as possible.

Discussion started in dev@airflow.apache.org

eladkal:
I think we can close this PR in favor of https://github.com/apache/airflow/pull/38111 ?

PR #960, opened 2024-02-17
Fix blogpost search

aritra24:
Blogpost search failed because certain articles didn't have tags which are assumed to exist leading to NPE. Also the hyperlink setting query selector had a type.


potiuk (approved):

aritra24:
Fixed the lint check failure

PR #977, opened 2024-03-06
Update use cases page

kentdanas:
We wanted to add some detail to this page on general use cases Airflow is used for in addition to the case studies. The new use case boxes map to the most common use cases reported in the Airflow survey, and each contain a description and a video talking through an example of the use case, link to a public repo with the example code, a "why Airflow?" section, and feature highlights with links to documentation.

The case studies are now called "testimonials", and are still on the same page.


ashb (approved):
Oh awesome. That makes it so much nicer.

LGTM me, I'll let a few more people weigh in first.

(I haven't proof read it, only skimmed)

potiuk:
This looks cool, I already made my comments to the content of the videos and voice-over (being a professional lector I pay attention to some quality problems ;) - and it's corrected now (thanks @kentdanas and @TJaniF !)

It's cool and nicely organised. Also, I noticed a bit too much "Astronomer" in the original videos - there were tabs visible and prominent Astronomer references. What's left is pretty ok for me, we have a mixture of "pure OSS" videos and few with slightly visible Astronomer reference, but the level it is now, is pretty ok for me. 

I really love the discoverability of cases it gives and the way that we are trying to reach out using various media. 

One thing I am **slightly** worried about is the weight it adds to our website to embed those videos in our git. This adds significant weight to it and we know we already have huge problems in `airflow-site` size and it might break our CI process even more eventually (at least until we possibly split out airflow-site) cc: @utkarsharma2.

Maybe it's a good idea to upload the videos to youtube and host them from there ?

kentdanas:
> One thing I am **slightly** worried about is the weight it adds to our website to embed those videos in our git. This adds significant weight to it and we know we already have huge problems in `airflow-site` size and it might break our CI process even more eventually (at least until we possibly split out airflow-site) cc: @utkarsharma2.
> 
> Maybe it's a good idea to upload the videos to youtube and host them from there ?

@potiuk this makes sense, we can definitely host them on YT. I think it makes the most sense to do this on the Apache Airflow YT (https://www.youtube.com/@ApacheAirflow), do you know who has access to post videos to that channel?

potiuk:
> @potiuk this makes sense, we can definitely host them on YT. I think it makes the most sense to do this on the Apache Airflow YT (https://www.youtube.com/@ApacheAirflow), do you know who has access to post videos to that channel?

Dicussion in slack :)

potiuk:
Looks good now.

PR #991, opened 2024-03-27
Remove Code of Conduct from the main menu

jedcunningham:
We don't need the Code of Conduct to be the first menu item in our main menu :)


o-nikolas (commented):
What position will it fall down to now? Or will it be removed entirely from the TOC at the top?

jedcunningham:
Removed completely. It already exists in the bottom menu, which is where it was intended to live. (Sorry, should have added that in the main description)

o-nikolas (approved):

PR #993, opened 2024-03-27
Add new meetup groups

Briana-Okyere:
Add 3 Meetup Groups to the Airflow site: Lisbon, Seoul, and São Paulo. Also alphabetized them.


potiuk:
Fixed `Sao Paulo` and added Warsaw

PR #1002, opened 2024-04-09
[Landing pages] Replace use case video iframes with links 

merobi-hub:
Embedding videos from external sources, as we are trying to do with the use case videos, is frowned upon by the ASF due to concerns about violating the GDPR. Currently, the videos are broken because the iframes violate the site's CSP, which is probably coming from ASF infra. Fixing the iframes, even if we could, seems ill-advised and could take time.

This replaces the iframes with links employing `img` tags. The appearance of the pages is unchanged.


potiuk (approved):

potiuk:
proably best choice

PR #1012, opened 2024-05-10
Add Scarf Pixel to Airflow RTD Theme

kaxil:
Similar to https://github.com/apache/superset/pull/25065 and https://github.com/apache/airflow/pull/39510 but this one doesn't need to wait for the VOTE since this adds a simple transparent tracking pixel. More detail in https://docs.scarf.sh/web-traffic/

All of this data will be available in Scarf, creds for which are shared to 1password for PMC members, and can be reported periodically in things like Town Hall or newsletters.


kaxil:
Once this is merged, will have to publish the new version to PyPI

jscheffl (approved):

PR #1038, opened 2024-07-14
Even more Airflow(TM) replacements to Apache Airflow(R)

potiuk:
There were few other places where "Airflow(TM)" rather than "Apache Airflow(TM)" was used - all that should be changed to "Apache Airflow(R)" after registration of the trademark.


hussein-awala (approved):
Approve blindly 🤞 
Screenshot 2024-07-14 at 23 53 29

PR #1060, opened 2024-08-22
Fix helm chart 1.15.0 release date

eladkal:
Version 1.15.0 was released on `2024-07-24` not on `2022-07-24`

https://github.com/apache/airflow/pull/41006 fixed it in the source but since doc were already generated we need regenerate the doc build for the site with manual fix.

### DIFF
--- a/docs-archive/helm-chart/1.15.0/release_notes.html
+++ b/docs-archive/helm-chart/1.15.0/release_notes.html
@@ -42,9 +42,9 @@
     </script>
     <!-- End Matomo Code -->
 
-    
+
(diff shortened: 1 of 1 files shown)


mg-aik (approved):

potiuk (approved):

eladkal:
> There are some whitespace changes - but it does not really matter.

I don't know why it does that in every change

potiuk:
> > There are some whitespace changes - but it does not really matter.
> 
> I don't know why it does that in every change

It's your IDE reformatting it

Issue #1008, opened 2024-04-20
Privacy menu item should link to https://privacy.apache.org/policies/privacy-policy-public.html


sebbASF:
The Privacy menu item should link to https://privacy.apache.org/policies/privacy-policy-public.html:

https://www.apache.org/foundation/marks/pmcs.html#navigation

This should be a direct link, rather than via a separate page as at present.

divyaguduru:
I would like to work on this issue .. can you pls assign it to me @sebbASF

pierrejeambrun:
@gdivya6028 assigned :)

Satishmedarr:
Hi sir, I would like to work on this issue. Please assign it to me. I'll ensure to add a visually appealing hover effect to the social icons in the footer. Thank you!

Satishmedarr:
Hi sir, I would like to work on this issue. Please assign it to me. I will update the Privacy menu item to link to the specified URL: https://privacy.apache.org/policies/privacy-policy-public.html. Thank you!

gdivya6028:
hi @potiuk , this issue can be closed as per our discussion on PR [https://github.com/apache/airflow-site/pull/1066]

PR #1066, opened 2024-09-07
Fix privacy notice to open apache website privacy policy page on click #1008

gdivya6028:
Fix privacy notice to open apache website privacy policy page on click #1008 

The Privacy menu item should link to https://privacy.apache.org/policies/privacy-policy-public.html:

This should be a direct link, rather than via a separate page as at present.


pierrejeambrun:
Should we also remove the `/privacy-notice/` landing page ?  (the original page that we are not using anymore)

gdivya6028:
@pierrejeambrun removed /privacy-notice/ related page changes now

pierrejeambrun:
Thanks for the update.

Do you mind explaining the motivation behind this change ? Is that an apache foundation requirement or more of user convenience to save a clicks ?

Indeed for consistency we will have to adapt the archives as well, for past build we can directly update the archives. You can do that manually or with a script.

For future builds you can check airflow repo code [here](https://github.com/apache/airflow/blob/main/docs/apache-airflow/privacy_notice.rst) and adapt the documentation accordingly.

potiuk:
Yes. I would also like to hear what is the motivation behind the change. The nice thing about the privacy policy being described in a separate page is that we can add our soecific things there where needed and that users have "airflow" link they can refer to, for example if needs be we might refert to and explain some specific decision choices we made - for example using scarf (which we probably should add to that page @kaxil BTW) . It's going to be better searchable also when users will search for "Airflow privacy policy" this way.

So having a separate page is good to have, I am not sure why we might want to replace it with direct link.

gdivya6028:
@pierrejeambrun / @potiuk , w.r.t quicklinks in footer screen, this fix is required to keep the user experience consistent for the quick links in the footer screen. An additional Interim landing page to navigate to Apache Privacy policy page is not necessary as it's redundant and only leads to extra clicks by the user. However, we can retain this interim landing page - `/privacy-notice/` , incase we are going to update with additional info that are not available in the Apache Privacy policy page currently.

In Airflow documentation page, we have sticky menu defined for the Airflow documentation page for easy navigation and the documentation page doesn't use the `/privacy-notice/` landing page code removed in this PR.. these menu items defined on the sticky menu in Airflow documentation page act as reference to the content on the screen and enables easier navigation to the user avoiding long scroll. The removal of Privacy policy content from the documentation page will also lead to the removal of Privacy policy from the sticky menu. In this context here, displaying Privacy Policy as a menu item on the sticky navbar that opens an interim page displaying the link to navigate to Apache Privacy Policy screen is totally fine I believe.

potiuk:
> incase we are going to update with additional info that are currently not available in the Apache Privacy policy page.

Yes. We want to keep it for that purpose, and we want to have all footer links to link to it - like today.

>  In this context here, displaying Privacy Policy as a menu item on the sticky navbar that displays the link to navigate to Apache Privacy Policy screen is totally fine I believe.

So, I understand we should just close that one? Everything is as expected currently? Or is there any fix you think is needed?

gdivya6028:
agree @potiuk , we are good to close this issue then.

Issue #1074, opened 2024-10-07
Main page shows "Apache Airflow&#174;" instead of "Apache Airflow®"


noamraph:
See those sections:
![image](https://github.com/user-attachments/assets/af795597-5e13-4a45-a323-7fb76d35933b)
It shows  `Apache Airflow&#174;` instead of `Apache Airflow®`

pierrejeambrun:
@noamraph Thanks for reporting that, it would be great if you could open a PR to fix that, it would really help :)

PR #1075, opened 2024-10-07
Fix ® symbol in landing page

noamraph:
Currently in two places in the landing page, `&#174;` appears instead of "®". See https://github.com/apache/airflow-site/issues/1074. Perhaps putting the actual Unicode character will fix it.


potiuk (approved):

potiuk:
Nice! Thank you :)

PR #1103, opened 2025-01-10
[BREAKING CHANGE] Update link tag for compatibility with Sphinx v7

shahar1:
This is a required change for upgrading Sphinx to version 7+ (see: https://github.com/readthedocs/sphinx_rtd_theme/issues/1465).
It won't work with older versions.

### DIFF
--- a/sphinx_airflow_theme/sphinx_airflow_theme/layout.html
+++ b/sphinx_airflow_theme/sphinx_airflow_theme/layout.html
@@ -223,7 +223,7 @@
 {%- endmacro %}
 
 {%- macro css() %}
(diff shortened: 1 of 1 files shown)


kaxil (approved):

potiuk (approved):

potiuk:
Ah... Did I merge too fast :) ?

potiuk:
Now - we need to do everything to merge the sphinx 7 changes before we build docs ;)

potiuk:
Ah  no... I see we should publish new theme for that . Let me do it

PR #1104, opened 2025-01-12
Bump theme version

potiuk:


potiuk:
I have big troubles in building the theme locally - but I think the packages are published as artifacts here and I should be able to download it from here and upload to PyPI

potiuk:
However. I am not sure what's going to happen if we publish it together with the old HTML code for past versions :) ?

shahar1 (approved):

shahar1:
> However. I am not sure what's going to happen if we publish it together with the old HTML code for past versions :) ?

Good point, because we don't pin the version in past evrsions.
Maybe we should create a completely new package just in case?

potiuk:
> > However. I am not sure what's going to happen if we publish it together with the old HTML code for past versions :) ?
> 
> Good point, past versions will get broken as we don't pin the version in past versions. Maybe we should create a completely new package just in case?

We can set "sphinx-airflow-theme>=0.0.12, <0.1.0" in `v2-10-test` and `providers-fab/v1-5`.

potiuk:
Though I **think** `css` is shared so the css should have support for both old and new formatting (cc: @jscheffl )

Issue #49, opened 2019-10-07
Implement autorefresh for Meetup's event


mik-laj:

somraj-2024:
Hey i can implement the auto refresh for the Meetup page. Please assign this issue to me.

o-nikolas:
@somraj-2024 done

o-nikolas:
> [@o-nikolas](https://github.com/o-nikolas) I have been working on setting up the code, and it took me some time to get everything in place. I have implemented an auto-refresh feature for the Meetup Events section to ensure the latest events are always displayed. Currently, I have set the refresh interval to 30 seconds. However, I wanted to check with you—what would be an ideal time interval for refreshing the events to balance performance and real-time updates? Looking forward to your guidance.

Hey @somraj-2024! Thanks for working on this. I think 30s is a bit too frequent, I wouldn't expect Meetup data to be changing on a such a frequent basis. But you should double check with @mik-laj who opened the issue. They provided no description for the task, but at least may have a better intuition about the frequency.

potiuk:
Yeah. I'd say 30 hours would be more appropriate :)

prashant2007-wq:
Hi maintainers,
I would like to work on this issue.

From the title, I understand that the Meetup events section should refresh automatically instead of depending only on static/manual updates. I will first inspect how Meetup/community events are currently fetched, stored, and rendered in the Airflow site.

### Approach

My plan is to keep the change focused and minimal:

- Locate the current Meetup/events data source and rendering flow.
- Check whether events are currently hardcoded, generated at build time, or fetched from an external source.
- Add an auto-refresh mechanism that updates Meetup events automatically during the site build or through the existing data-generation workflow.
- Preserve the current UI and layout of the events section.
- Add safe fallback behavior so the page still builds if the Meetup/events source is unavailable.
- Avoid exposing secrets or API keys in the frontend.
- Document how the refresh flow works if a new script or data-generation step is added.
- Test the site build locally and verify that the events section still renders correctly.

Please let me know if this implementation direction matches what you expect for this issue, or if there is a preferred data source/API that should be used for Meetup events.

potiuk:
Figuring out how to do it is part of the task - so come up with a proposed PR.

PR #1128, opened 2025-02-16
Implemented auto refresh for Meetups Event 

somraj-2024:
For each 30 hrs, the Meetup Events will refresh automatically.


potiuk (changes requested):
Why do we have such big changes in .lock files

Also I start to question why would we need it all. Isn't that simply refreshed where the meetup page is refreshed? why auto-refresh is needed ? Do we expect people to look at the meetup page and wait for it to refresh?

somraj-2024:
This PR is related to an issue which was raised earlier [Implement auto refresh for Meetups Event](https://github.com/apache/airflow-site/issues/49) Also, we had a discussion about it in comments.

About the .lock files, I haven't made any changes in those files, other than the meetup.js file. During the code setup and installation of the packages, some files might have changed. I will fix it.

PR #1145, opened 2025-03-24
Update _index.html

OGRIN0:
add it


potiuk:
Please avoid such changes on public Airflow repo. It's not a testing playground and adds noise for maintainers.

PR #1146, opened 2025-03-25
Suggest changes link update for airflow 2.10.5 Docs

amolsr:
Closes [#48178](https://github.com/apache/airflow/issues/48178)


potiuk:
Thank you !

amolsr:
hi @potiuk , I tried to build the site but it feels like the Hugo 0.91.2 is not compatible with MacOS. 
Screenshot 2025-03-26 at 12 22 18 AM

amoghrajesh (approved):
Looks OK! If there are any issues in the future, we will fix in follow ups. Thank you for doing the hard work

PR #1161, opened 2025-04-16
Rewrote site.sh to site.py

NarmalaSk:
This pull request introduces the following changes:

Python Script: I have replaced the existing site.sh with a Python script (site.py) using argparse for argument parsing, which improves flexibility and maintainability.

Reason for Change: I addressed issues with outdated libraries and configurations in the Node.js Airflow site during the build process.

Workflow Update: I have updated the GitHub Actions workflows to use site.py instead of site.sh.

Future Plans:

I will start the migration of the Airflow site to the Pelican static site builder in a separate branch.

After testing, I will work towards replacing the current site with the Python-based static site.

If the Python static site is in already development, I will contribute to it and refactor site.py for compatibility.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -77,7 +77,7 @@ jobs:
           extended: true
       - name: 🟢 Install node dependencies
         run: |
-          /mnt/airflow-site/site.sh install-node-deps
+          python3 /mnt/airflow-site/site.py install-node-deps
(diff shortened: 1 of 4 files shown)


NarmalaSk:
Also we can change the site.py accordingly for new build using getpelican ,But iam unsure about the sphinx theme

PR #1162, opened 2025-04-17
Hide the "Suggest a change" button when TOC is hovered

dion-ricky:
I'm unable to test locally, looks like there's something wrong with my setup. I followed the README but when I tried to install the sphinx theme, the required files (css and js) was not found. Therefore, I tested it by adding the class manually to the Chrome DevTools and interacting with the elements directly in my browser. At least from my test it is working as expected. If you can lend a hand to help me test my changes, I would really appreciate it.

### DIFF
--- a/landing-pages/site/assets/scss/_suggest-change.scss
+++ b/landing-pages/site/assets/scss/_suggest-change.scss
@@ -1,0 +1,13 @@
+.fade-target {
+    opacity: 1;
+    visibility: visible;
+    pointer-events: auto;
+    transition: opacity 0.5s ease, visibility 0s linear;
+}
+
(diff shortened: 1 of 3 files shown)


dion-ricky:
Adjusted according to pre-commit check

potiuk:
Cool. Thanks. I will have to release and rebuild airflow site docs with it so this will take a bit to fix it - I have another theme change to fix together with the "suggest" button so I am on it.

potiuk:
Works nicely - tested it locally - we will re-publish Airflow docs soon to get it done.

potiuk:
You can check in about  an hour after shift-refresh at airlfow site.

PR #1170, opened 2025-04-23
Fix verison replacement bug in 3.0.0 docs

potiuk:


potiuk:
Yeah. One dict missing

PR #1174, opened 2025-04-24
Update suggest links in older versions

potiuk:


potiuk:
BTW. I've done that in the way that old links are going against `v2-10-test` branch. especially that we had a number of files moved and removed here and there - but if someone opens it - at least people will not get 404 and we can always ask them to change it to main.

PR #1181, opened 2025-05-01
Update metadata gen and add RewriteRule

gopidesupavan:
With new docs publishing from S3, we will be going to use AWS CloudFront in .htaccess file to proxy. all the docs will be served from S3, except landing pages.

There couple of PR's followed up after this merge.

https://github.com/apache/airflow-site/pull/1182

https://github.com/apache/airflow-site/pull/1184

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -107,6 +107,30 @@ jobs:
           rm -rf ./docs-archive/*
           echo "docs-archive directory files are deleted as they are no longer needed after the site is built in the above step"
           /mnt/airflow-site/.github/scripts/print_df.sh
+
+      - name: Install AWS CLI v2
+        if: ${{ github.event_name == 'push' }}
+        run: |
+          curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o /tmp/awscliv2.zip
+          unzip -q /tmp/awscliv2.zip -d /tmp
+          rm /tmp/awscliv2.zip
+          sudo /tmp/aws/install --update
+          rm -rf /tmp/aws/
+
+      - name: Configure AWS credentials
+        if: ${{ github.event_name == 'push' }}
+        uses: aws-actions/configure-aws-credentials@010d0da01d0b5a38af31e9c3470dbfdabdecca3a  # v4.0.1
+        with:
+          aws-access-key-id: ${{ secrets.DOCS_AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.DOCS_AWS_SECRET_ACCESS_KEY }}
+          aws-region: us-east-2
+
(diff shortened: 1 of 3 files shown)


landing-pages/site/static/.htaccess
```diff
@@ -1,3 +1,8 @@
+RewriteEngine On
+
+RewriteCond %{REQUEST_URI} ^/docs [NC]
+RewriteRule (.*) http://staging-docs-airflow-apache-org.s3-website.us-east-2.amazonaws.com/$1 [P]
```
gopidesupavan, line 4: We can change this url to live docs s3, once we are good to go.

gopidesupavan:
Alright lets keep it ready to go.

Once we switch we may have to update github workflows on this repo to remove not required steps.

landing-pages/site/static/.htaccess
potiuk: Nice.

.github/workflows/build.yml
```diff
@@ -107,6 +107,30 @@ jobs:
...
+        with:
+          aws-access-key-id: ${{ secrets.DOCS_AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.DOCS_AWS_SECRET_ACCESS_KEY }}
+          aws-region: us-east-2
+
+      - name: Copy docs index page to S3 docs
+        if: ${{ github.event_name == 'push' }}
+        run: |
+          aws s3 cp ./dist/docs/index.html s3://staging-docs-airflow-apache-org/docs/index.html
+          aws s3 cp ./dist/docs/index.xml s3://staging-docs-airflow-apache-org/docs/index.xml
```
gopidesupavan, line 132: The index.html file generated in this repository is required at the S3 location; otherwise, we encounter a "Not Found" error. Therefore, the only solution i think is to copy this file to S3 as part of the generation process.

I attempted to use a redirect, but it resulted in an infinite loop—this behavior is expected since we're proxying through S3.

potiuk:
BTW. Question: do we have cloudfront setup for the s3 bucket? I guess we'd need that otherwise that might generate a lot of cost and be a bit slow ?

gopidesupavan:
> BTW. Question: do we have cloudfront setup for the s3 bucket? I guess we'd need that otherwise that might generate a lot of cost and be a bit slow for people from various regions in the world?

Good question you raised, we have to setup that aswell.

gopidesupavan:
> > BTW. Question: do we have cloudfront setup for the s3 bucket? I guess we'd need that otherwise that might generate a lot of cost and be a bit slow for people from various regions in the world?
> 
> Good question you raised, we have to setup that aswell.

Setuped now..

potiuk (approved):
NICE!

PR #1182, opened 2025-05-05
Update new airflow site docs publishing process

gopidesupavan:


gopidesupavan:
@potiuk as you know already staging website useful for testing, and its a different bucket, so we have to create another CloudFront URL, should we create one for staging or we can also use s3 static website url ?, offcourse we dont use much once we are settle, on the live site. so less traffic on staging site. Any suggestion.

gopidesupavan:
@eladkal i have documented here some steps for the new process, its still in draft, but some steps will help you to run the publishing docs to s3.  

I am yet to add steps for testing docs without publishing docs to live site.

PR #1188, opened 2025-05-08
Remove protected-branches for now

potiuk:


potiuk:
I need that in order to be able to clean-up history

PR #1192, opened 2025-05-09
Add workflow_dispatch to docs build

gopidesupavan:
This will help to refresh airflow-site when new versions published to S3. No empty commit is required to trigger.


gopidesupavan:
I think we can automate this aswell, once docs published to S3, we can trigger this workflow from airflow with github api.

potiuk (approved):
Yep

PR #1195, opened 2025-05-11
Modernize GH action environment setup

potiuk:


kaxil (approved):

potiuk:
Hmm. That one is strange: 

WARN 2025/05/11 13:38:55 found no layout file for "HTML" for layout "search" for kind "page": You should create a template file which matches Hugo Layouts Lookup Rules for this combination.

potiuk:
I think I am using same versions of node/js/hugo as previously :thinking:

potiuk:
Ah ... submodules are needed :facepalm:

potiuk:
All right - got it working :)

gopidesupavan:
Nice :)

PR #1196, opened 2025-05-11
Switch sphinx_airflow_theme build to pyproject.toml / uv

potiuk:


potiuk:
Ok. Produced .whl has the same content as previous one

PR #1202, opened 2025-05-16
Use correct bucket to metadata file

gopidesupavan:


gopidesupavan:
The same change needs on staging branch. as it currently points to live.

potiuk (approved):

potiuk:
Nice!

potiuk:
> The same change needs on staging branch. as it currently points to live.

We can just rebase `staging` branch on top of `main` -> It is designed to be done this way.

gopidesupavan:
> > The same change needs on staging branch. as it currently points to live.
> 
> We can just rebase `staging` branch on top of `main` -> It is designed to be done this way.

ah okay :)

potiuk:
I rebased `staging`

Generatlly what I do:

```bash
git checkout main
git pull
git branch -f staging
git push origin staging
```

Unless we develop something specific in staging - staging is generally simply behind main and can be fast-forwarded like that.  If we diverge in staging, it will fail at the last stage because it will not be fast-forwardeable.

PR #1210, opened 2025-06-23
Quote descriptions with special characters in chart index file

jedcunningham:
Artifacthub is complaining about these - unfortunately, I'm not sure if its complaining from the index file or the entry in Chart.yaml. One way to find out...


jedcunningham:
And... its from Chart.yaml :(

PR #1216, opened 2025-07-04
Better index page for "Documentation".

potiuk:
* Moving Docker Stack, Helm Chart and Python Client above providers (on most of the screens you will still see providers and realise you should scroll down to see all of them.

* Removing Go Client - we do no release it any more.


gopidesupavan:
cool indeed good

PR #1218, opened 2025-07-07
Removing DoubleCloud from 'Airflow as a Service'

areeb1501:
Double cloud has wound down operations. Would be better to remove this from the official airflow ecosystem page to avoid misleading URLs. As of October 2023, new customers are no longer allowed to signup.


potiuk:
Thanks!

potiuk (approved):

PR #1224, opened 2025-07-22
Do not rewrite ^docs/index.html on our website

potiuk:
The docs/index.html is generated and published in the ASF infrastructure and we should not rewrite it and redirect to our S3 - this way whatever is the latest version on our "airflow-site" is the one published.

So far, we copied the index over to S3 when we built documentation,
but that step is not really needed any more if we do not rewrite the
docs/index* URLS. This is more stable and "proper".
    
We also do not have to optimize the artifacts any more - airflow-site
does not produce huge website any more, so we do not have to
delete the generated docs/index.html because there is no archive there
any more.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -27,7 +27,6 @@ jobs:
   build:
     runs-on: ["ubuntu-22.04"]
     env:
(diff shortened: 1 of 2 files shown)


potiuk:
Let's see

gopidesupavan:
Nice, was added this when i tested it was not redirected properly so copied the file :) nice update.

potiuk:
Yeah. I am still not sure WHY the S3 version of it was overwritten by older version though :). But this one seem to work nicely.

PR #1229, opened 2025-08-14
Serve /schemas/ from the S3 bucket/Cloudfront too.

ashb:
This gives us a place outside of versioned docs to Publish Open API or JSON
schemas.

I removed the `NC` flag which makes it do a case-insensitve comparission as
the case actually matters here, if we have a request to `/Docs` it will 404,
so there's no point sending it to S3

This is #1228 but for the staging branch too


ashb:
Not needed

PR #1233, opened 2025-08-25
Fix #1220 missing scrollbar in _dropdown.scss

### LINKED ISSUE #1220: Cannot scroll to view older versions (e.g., 2.5.1) in docs version selector
On the Airflow documentation site, the version selector dropdown doesn’t support scrolling.
As a result, it's not possible to access older versions like 2.5.1 and earlier.

https://github.com/user-attachments/assets/b0f4d11d-4043-4d55-a39a-7d036ccd56a5

it can fix this by adding the following CSS to the .dropdown-menu:
```
max-height: 300px;
overflow-y: auto;
```
This will make the dropdown scrollable and allow users to view all available versions.

camiloll:
Fix dropdown in docs-version-selector not displaying a scrollbar to select older versions

### DIFF
--- a/landing-pages/site/assets/scss/_dropdown.scss
+++ b/landing-pages/site/assets/scss/_dropdown.scss
@@ -33,4 +33,9 @@
   a {
     @extend .bodytext__medium--brownish-grey;
   }
+
+  .dropdown-menu {
+    max-height: 100%;
+    overflow-y: auto;
+  }
 }


potiuk:
Let's see. Rebuilding the site to see ;)

potiuk:
We'll see it when the next version of docs is generated and published.

PR #1263, opened 2025-10-27
Fix Main animation not displaying

choo121600:
Fixed an issue where the header animation was not rendered due to an incorrect p5 import.

## Before
image

## Now
https://github.com/user-attachments/assets/cbc652a5-74cf-4408-ba38-d5a5cfada733


choo121600:
fixed in #1266

PR #1284, opened 2025-11-23
Improve footer hover interaction in documentation footer

alok-kumar0421:
### What does this PR do?
This PR improves the hover interaction for footer links in the Airflow documentation website.

### What was the issue?
Some footer elements (social links, policy links, and text links) did not show any hover feedback.  
This reduced accessibility and made the footer feel unresponsive.

### What has been changed?
- Added smooth hover animation and opacity transition to footer media links.
- Added hover color change for policy-item links.
- Added cursor pointer to interactive footer elements.
- No layout or structural changes—only visual interaction improvements.

### Why is this important?
This enhances the UX, makes clickable elements more obvious, and improves consistency across the site.

### Testing
Tested locally by building the docs.  
All styles render correctly on:
- Desktop
- Tablet
- Mobile layouts

This change does not affect any Python code or documentation content—CSS only.

### DIFF
--- a/landing-pages/site/assets/scss/_footer.scss
+++ b/landing-pages/site/assets/scss/_footer.scss
@@ -37,6 +37,13 @@ footer {
 
       &--link {
         margin-right: 30px;
+        cursor: pointer;
+        transition: opacity 0.2s ease, transform 0.2s ease;
+
+        &:hover {
+          opacity: 0.7;
+          transform: scale(1.05);
+        }
       }
 
       &--button-with-text {
@@ -64,6 +71,9 @@ footer {
       }
 
       &--policy-item {
+        cursor: pointer;
+        transition: opacity 0.2s ease, color 0.2s ease;
+
         &::before {
           content: "\00a0\00a0";
         }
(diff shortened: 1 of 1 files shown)


potiuk:
Nice. Tested it on https://airflow.staged.apache.org/ and it's way better :)

potiuk:
Thanks for this small QOL improvement - it should be live in ~ 5 minutes (you will need Shift-Refresh to speed up browser caching).

alok-kumar0421:
Thank you for reviewing and merging the PR!  
I appreciate the feedback.  
I’d be happy to contribute more if there are any other improvements needed.

potiuk:
Cool @alok-kumar0421 

I think we wanted to make the "Version drop-down" to be better. This needs quite a bit more CSS/JS interactions - maybe you can take a look? 

Currently we are limiting the versions shown to the list of "latest" minor versions - but it would be great if the version selector allows not only to choose from the drop-down list, but also be able to type-in the version (say 3.1.0) - currently to get to the version you need to replace the version in the URL, but the version selector could be more of a "combo-drop-down" - where you can both select and type the version (bonus points if the version could be auto-complete'able but that would require some back-end changes - currently the list of versions only contains the latest minor versions).

alok-kumar0421:
Thanks for the suggestion!  
Yes, I can take a look at improving the "Version drop-down" selector.

I'll explore how to:
- allow typing a version (combo input),
- keep the dropdown selection working,
- and possibly support autocomplete if feasible on the frontend.

I'll share a plan or initial approach soon.

PR #1296, opened 2025-11-28
Fix Format release links for Apache Airflow announcements

choo121600:
Updated links to include line breaks for better readability.

## before
image

## after
image


potiuk:
Added fixup with upgrade check.

potiuk:
Thanks!

PR #1314, opened 2025-12-12
Fix Previous/Next button navigation in use cases

versha890:
This PR fixes the unintuitive behavior of the "Previous" and "Next" buttons on the Use Cases pages of the Airflow website.

**Problem:**
- "Previous" navigated forward (to the next item)
- "Next" navigated backward (to the previous item)

**Solution:**
- Swapped `.PrevInSection` and `.NextInSection` in `layouts/partials/pager.html`
- Now "Next" → right, "Previous" → left, providing intuitive navigation

**Testing:**
1. Run `hugo server -D` locally
2. Open http://localhost:1313/use-cases/etl_analytics/
3. Verify buttons navigate correctly

### DIFF
--- a/landing-pages/site/layouts/partials/pager.html
+++ b/landing-pages/site/layouts/partials/pager.html
@@ -25,5 +25,3 @@
         {{ partial "buttons/button-hollow" (dict "text" "Next" "disabled" (not .PrevInSection)) }}
     </a>
 </div>
-
-</nav>


landing-pages/site/layouts/partials/pager.html
```diff
@@ -18,10 +18,12 @@
...
+        {{ partial "buttons/button-hollow" (dict "text" "Previous" "disabled" (not .NextInSection)) }}
     </a>
-    <a {{if .NextInSection}}href="{{.NextInSection.RelPermalink}}"{{end}}>
-        {{ partial "buttons/button-hollow" (dict "text" "Next" "disabled" (not .NextInSection)) }}
+    <a {{ if .PrevInSection}}href="{{.PrevInSection.RelPermalink}}"{{end}}>
+        {{ partial "buttons/button-hollow" (dict "text" "Next" "disabled" (not .PrevInSection)) }}
     </a>
 </div>
+
+</nav>
```
choo121600:
```suggestion
</div>

```

choo121600: Good Catch 🙌
However, a small change is needed. The `</nav>` tag is not necessary.

Issue #1312, opened 2025-12-12
UI inconsistency in nav due to extra margin.


SHASHI9705:
Due to the extra padding in top and bottom of the nav the selected heading to read is geting hidden with some of their content and user have to scroll again to read properly .

https://github.com/user-attachments/assets/e0c386dc-6169-4924-97b2-e0fb0ffcd548

And this improvemnet need for the other some pages also so proposed fix is to have some less padding in the nav and a bit of margin in top of the content.

@potiuk  Kindly verify and assigne it to me , i'd like to work on this issue .

potiuk:
feel free - again, just PR is enough

SHASHI9705:
@potiuk i have made a pr for this, kindly review and merge it.

PR #1316, opened 2025-12-12
Padding of Nav.

SHASHI9705:
Decreased the padding of Nav to get good consistency for the issue #1312 
@potiuk

### DIFF
--- a/landing-pages/site/assets/scss/_navbar.scss
+++ b/landing-pages/site/assets/scss/_navbar.scss
@@ -27,7 +27,7 @@
   justify-content: flex-start;
   border-bottom: solid 1px map-get($colors, very-light-pink);
   z-index: 32;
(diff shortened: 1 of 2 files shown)


choo121600:
For easier review, could you please upload a screenshot?

choo121600:
However, I’m not entirely convinced by the concern that the current padding causes the content to be obscured.
I do agree that the header height is relatively large, but as far as I know, there is currently no part of the content that is actually hidden.
If there is any area where the content is obscured, could you please share a screenshot?

choo121600:
Also, if only the header padding is adjusted, it results in a noticeably larger gap between the header and the main content, as shown below.
In that case, I think the height (or layout) of the main content would need to be adjusted as well.

## Now
image

image

## Your Change
image

PR #1318, opened 2025-12-13
Fix inconsistent hover and underline behavior on Community page links

SHASHI9705:
### What does this PR do?

Fixes inconsistent hover and underline behavior on links in the Community page,
ensuring a consistent visual experience across all community link items.

### Why is this needed?

Currently, some Community page links show different underline or hover effects,
which feels visually inconsistent and distracting.

@potiuk

PR #1270, opened 2025-11-07
Upgrade ancient Docsy theme and implement dark mode

kaxil:
**Preview**: 
- Homepage: https://airflow-dark.surge.sh/ 
- Airflow Core: https://airflow-dark.surge.sh/docs/apache-airflow/stable/

This PR upgrades the Docsy theme from v0.2.0-pre (from 2019 !!) to v0.12.0 (latest) and implements Bootstrap 5 dark mode support for both the landing pages and Sphinx documentation. Used https://www.docsy.dev/docs/adding-content/lookandfeel/ and https://www.docsy.dev/docs/updating/ for Upgrading -- it wasn't straight-forward unfortunately but I think I have covered everything that I could find/identify. (including Confetti :D )

### Key Changes

* **Theme Upgrade**: Docsy v0.2.0-pre → v0.12.0
* **Bootstrap**: Upgraded from Bootstrap 4 to Bootstrap 5.3.3
* **Dark Mode**: Added theme toggle with light/dark/auto modes

### Visual Changes

image

image
image
image
image
image
image
image
image

#### Landing Pages (Hugo/Docsy)
- Enable `showLightDarkModeMenu = true` in config.toml
- Add theme toggle component to navbar (light/dark/auto modes)
- Store user preference in localStorage
- Use Bootstrap 5 CSS variables for all UI elements

#### Sphinx Documentation
- Add Bootstrap 5 theme toggle to header
- Style all elements (sidebar, tables, admonitions, forms) for dark mode
- Use Pygments-generated GitHub Dark theme for syntax highlighting

#### Technical Requirements
- Added `$td-sidebar-tree-root-color` variable (required by Docsy v0.2.0+)
- Added `$font-awesome-font-name` variable (required by Docsy v0.6.0+)
- Added `.-bg-orange` utility class (required by Docsy v0.6.0+)
- Created `pygments/_dark.scss` (generated via: `pygmentize -S github-dark`)
- Use `hooks/body-end.html` for webpack scripts (proper Docsy pattern)

### Testing

- [x] Verified dark mode toggle works on landing pages
- [x] Confirmed confetti animation adapts to theme changes
- [x] Tested all admonition types in light and dark modes
- [x] Verified code syntax highlighting in Sphinx documentation
- [x] Checked sidebar contrast in dark mode
- [x] Tested theme persistence across page navigation
- [x] Verified font consistency (Open Sans) across Hugo and Sphinx pages

PR #1325, opened 2025-12-14
Improve scroll-to-top button readability with icon-only UI

SHASHI9705:
### What does this PR do?

This PR improves the readability of the floating scroll-to-top button by removing
the visible text label and using an icon-only presentation.

The arrow icon already clearly conveys the button’s purpose, and the text label
was overlapping page content on longer pages.

Before Changes:

Screenshot 2025-12-14 180942

After Changes:

Screenshot 2025-12-14 180911

### Why is this change needed?

- The visible “Scroll to top” text overlaps content and reduces readability
- The icon alone is sufficient and follows common UI patterns

LGTM :) kindly verify and merg it .
@choo121600 @potiuk

### DIFF
--- a/landing-pages/site/assets/scss/_buttons.scss
+++ b/landing-pages/site/assets/scss/_buttons.scss
@@ -50,12 +50,6 @@ button {
       width: auto;
     }
 
-    @media (max-width: 1279px) {
-      svg {
-        padding-right: 15px;
-      }
-    }
-
     span {
       padding-left: 15px;
       display: inline-block;
--- a/landing-pages/site/assets/scss/_base-layout.scss
+++ b/landing-pages/site/assets/scss/_base-layout.scss
@@ -79,10 +79,6 @@
 
     &--scrollButton button svg {
       display: block;
-      margin: 0 auto;
-      position: relative;
-      left: 56%; // Move just a little bit more to the right
-      transform: translateX(-50%);
     }
 
   }


choo121600:
As mentioned https://github.com/apache/airflow-site/pull/1318#issuecomment-3649184426 ,
It’s better not to tag specific people. I’m just an individual contributor who loves the Apache Airflow project, like you, not a maintainer or committer. I’m only reviewing to help save the maintainers’ time, and I don’t have any permissions.

Even if you don’t tag anyone, others will still review your code.

landing-pages/site/assets/scss/_buttons.scss
choo121600, line 1: We need to keep padding-right: 15px;.
because of "suggest a change in this page" button

choo121600 (changes requested):
And I think there might be a better way to solve this!

landing-pages/site/assets/scss/_buttons.scss
SHASHI9705: The better idea is to put padding-left: 15px; in the span instead of svg , because it will show extra space even if there is no text present . I have updated that part now, LGTM :)

landing-pages/site/layouts/partials/buttons/button-with-arrow.html
choo121600, line 1: I think `button-with-arrow` is only used here(`landing-pages/site/layouts/partials/scroll-to-top.html`).
In that case, if we completely remove the `text` element there, we wouldn’t need the part that checks `if .text`, right?
of course, it would be good if you could double-check what I said. I haven’t looked at all the code yet, so I’m speaking based on what I remember.

choo121600 (changes requested):

landing-pages/site/assets/scss/_buttons.scss
choo121600: I think this is a meaningless change.

choo121600:
And please update the screenshots whenever there are changes in the code :)

landing-pages/site/assets/scss/_buttons.scss
SHASHI9705: This is not meaningless change, you just metioned that padding needed for the "suggest a chnage in this page" button so we can put padding in the span class instead of puting that in svg class, So that there will be no extra space in arrow button as there is no text present.

landing-pages/site/layouts/partials/buttons/button-with-arrow.html
SHASHI9705: yes may be that would be better .

SHASHI9705:
Before : 

Screenshot 2025-12-15 141938

Unecessary extra text and padding.

After :

Screenshot 2025-12-15 142104

No extra unecessary text in scroll button and padding in svg, Only padding in span class for the " Suggest a change on this page " Button.

LGTM :) for merging it, kindly review and merge it. @potiuk

potiuk (approved):

potiuk:
Please stop pinging me with every PR you make. This is not how it works here - you make a PR and **some** reviewer at **some** time will review things possibly  and it might take days. You seem to be super impatient and ping me to get this faster - this comes as a very pushy behaviour.

SHASHI9705:
> Please stop pinging me with every PR you make. This is not how it works here - you make a PR and **some** reviewer at **some** time will review things possibly and it might take days. You seem to be super impatient and ping me to get this faster - this comes as a very pushy behaviour.

Sorry about that, I understand now. I won’t ping again and will wait for the normal review process. Thanks for letting me know.

landing-pages/site/assets/scss/_buttons.scss
choo121600: Image

If you change it this way, it will cause issues in responsive layouts.
Once this part is resolved, it looks good :)

choo121600 (changes requested):
Image

SHASHI9705:
Completed the fix of responsiveness -

In large screen : 

Screenshot 2025-12-16 025425

In small screen : 

image

I think now it's look good . Thanks :)

landing-pages/site/assets/scss/_buttons.scss
```diff
@@ -48,10 +48,16 @@ button {
...
       height: $icon-height;
       width: auto;
-      padding-right: 15px;
+    }
+
+    @media (max-width: 1279px) {
+      svg {
+        padding-right: 15px;
+      }
     }
```
choo121600: I think it would be better to address the root cause of the issue first.

The root cause seems to be in the following part of the code:
https://github.com/apache/airflow-site/blob/48c9a3bed01af394ffde6ca9c573d3d1eaa9584c/landing-pages/site/assets/scss/_base-layout.scss#L80-L86

If we improve this section, we may be able to avoid adding additional media queries.

SHASHI9705: yes i have fixed the scroll button with the layout, and also removed the additional media queries. Thanks:)

potiuk:
Nice. Great watching some real CSS expert discussions :)

image

potiuk:
Hmm. I think there is a small problem the scroll-to-top change created:

image

This button "Suggest changes" used to be on the right side (see the previous versions of the pages) - I think the scroll-to-top changed that button to be displayed on the left side - which is not really what we would expect.

choo121600:
??!!! I’ll check it 🤔
image

SHASHI9705:
> Hmm. I think there is a small problem the scroll-to-top change created:
> 
> image
> This button "Suggest changes" used to be on the right side (see the previous versions of the pages) - I think the scroll-to-top changed that button to be displayed on the left side - which is not really what we would expect.

The buttons are seem to be at normal position as expected in my browser :

image
Screenshot 2025-12-16 175515

See these pages , scroll button on left and suggest button on right side.

SHASHI9705:
> Hmm. I think there is a small problem the scroll-to-top change created:
> 
> image
> This button "Suggest changes" used to be on the right side (see the previous versions of the pages) - I think the scroll-to-top changed that button to be displayed on the left side - which is not really what we would expect.

Yes, I believe this can be fixed so it aligns to the right as expected.
Should I open a new PR for this change? I’m happy to do so if that’s preferred.

potiuk:
This happens only in 3.1.5 (only 3.1.5 had the scroll-to-top changes applied when generated - this could be an early version of it though - or maybe some other recent change - but the position of the button suggest that the two are connected.

SHASHI9705:
> This happens only in 3.1.5 (only 3.1.5 had the scroll-to-top changes applied when generated - this could be an early version of it though - or maybe some other recent change - but the position of the button suggest that the two are connected.

Yes, i think both are connected by the button class but have a seprate classes also for their change only, we can fix the suggest button by changing the code in "suggest" button class.

potiuk:
> Should I open a new PR for this change? I’m happy to do so if that’s preferred.

Ideally :) .. 

And after we finish the round of improvements - and after @kaxil merges his Dark Mode changes (and I think it would be great if our CSS experts help with reviewing that one):   https://github.com/apache/airflow-site/pull/1270 we will have to re-synchronize the css /html changes in all the historical versions of airflow versioned docs - this way we might get scroll-to-top everywhere and dark mode everywhere etc.

I already have some experience with using Copilot + Claude Sonet to do it across 860.000 of our generated Html pages :) and will be happy to do it (this boils down to a really long conversation with AI agent and making it do some awfull things in the generated hundreds of thousands files)  but I think we need to get this round of UX improvements tested and fixed before I attempt to do it.

While it's some fun to do this conversation, it's not fun enough for me to do it too often :D

choo121600:
> Yes, I believe this can be fixed so it aligns to the right as expected.
> Should I open a new PR for this change? I’m happy to do so if that’s preferred.

I’m currently on vacation, so if you’ve found a solution, would you mind submitting it as a separate PR?😉

potiuk:
>  I’m currently on vacation, so if you’ve found a solution, would you mind submitting it as a separate PR?😉

Wait, what ? Aren't you supposed to be on vacation and NOT respond ;) ?

choo121600:
I’m working on a tablet right now, so it’s not very easy for me 😄

potiuk:
> I’m working on a tablet right now, so it’s not very easy for me 😄

Heh - at least you can tests some responsive design issues ;)

SHASHI9705:
> I’m currently on vacation, so if you’ve found a solution, would you mind submitting it as a separate PR?😉

Sure, no problem at all 😊

choo121600:
Coool🙌

SHASHI9705:
> Coool🙌

What if we put the scroll button also in that pages as all other pages are having that it will look good and match the layout , and also resolve the issue of "suggest" button , coz their class is having justify-content : space between;   puting the scroll button will resolve the issue and also match the layout of other pages.

potiuk:
> > Coool🙌
> 
> What if we put the scroll button also in that pages as all other pages are having that it will look good and match the layout , and also resolve the issue of "suggest" button , coz their class is having justify-content : space between; puting the scroll button will resolve the issue and also match the layout of other pages.

Yes that's what I am aiming at.

choo121600:
> > Coool🙌
> 
> What if we put the scroll button also in that pages as all other pages are having that it will look good and match the layout , and also resolve the issue of "suggest" button , coz their class is having justify-content : space between; puting the scroll button will resolve the issue and also match the layout of other pages.

Yes, I think we can solve it like this ;)

choo121600:
@SHASHI9705 
I’ve looked into this, and it seems the root cause lies in the following part.
There is a difference here compared to the landing page:

https://github.com/apache/airflow-site/blob/3c89f6f1aedb51d836463e3496c580f92c2b8887/sphinx_airflow_theme/sphinx_airflow_theme/layout.html#L464-L472

Unlike the Airflow-site landing page, the documentation pages are built using sphinx_airflow_theme.
aligning the implementation here with the approach used on the landing page should resolve the issue 😉

SHASHI9705:
> @SHASHI9705 I’ve looked into this, and it seems the root cause lies in the following part. There is a difference here compared to the landing page:
> 
> https://github.com/apache/airflow-site/blob/3c89f6f1aedb51d836463e3496c580f92c2b8887/sphinx_airflow_theme/sphinx_airflow_theme/layout.html#L464-L472
> 
> Unlike the Airflow-site landing page, the documentation pages are built using sphinx_airflow_theme. aligning the implementation here with the approach used on the landing page should resolve the issue 😉

Yes, i have checked the code and the root cause is that suggest button file is only include and that file is having just suggest button, we can  proceed by adding a scroll button to it , i'll open a pr in a while.

potiuk:
> Yes, i have checked the code and the root cause is that suggest button file is only include and that file is having just suggest button, we can proceed by adding a scroll button to it , i'll open a pr in a while.

Also as you are on it - after @kaxil  added the dark mode, the scroll-to-top button has disappearing arrow in dark mode while scrolling - so I guess a small adjustment should be needed to adapt the button to dark mode - and also "suggest a change" button is still the same (white) in dark mode, so likely we could address all of those in a single css update :)

kaxil:
Yeah I am doing some diagnostic/next-steps/fixups on staging too: https://airflow.staged.apache.org/docs/apache-airflow/3.1.5/best-practices.html (fyi)

SHASHI9705:
> > Yes, i have checked the code and the root cause is that suggest button file is only include and that file is having just suggest button, we can proceed by adding a scroll button to it , i'll open a pr in a while.
> 
> Also as you are on it - after @kaxil added the dark mode, the scroll-to-top button has disappearing arrow in dark mode while scrolling - so I guess a small adjustment should be needed to adapt the button to dark mode - and also "suggest a change" button is still the same (white) in dark mode, so likely we could address all of those in a single css update :)

yes i have already seen that and updated the code in the upcoming pr.

PR #1329, opened 2025-12-16
Fix OpenGraph modified_time to use .Params.lastmod

versha890:
This PR fixes the issue with the OpenGraph `article:modified_time` meta tag.  
Previously, regenerating the site updated the modification time incorrectly for all posts.  
Now, it uses `.Params.lastmod` to reflect the actual last modified date set in the post front matter.

##Changes
- Updated `_partials/opengraph.html`:
  ```go
  {{ with .Params.lastmod }}
    <meta property="article:modified_time" content="{{ . | time.Format "2006-01-02T15:04:05Z07:00" }}" />
  {{ end }}

### DIFF
--- a/landing-pages/site/assets/scss/_community-page.scss
+++ b/landing-pages/site/assets/scss/_community-page.scss
@@ -231,6 +231,3 @@
  .roadmap .wy-nav-side-toc {
    top: 263px !important;
  }
-
-
-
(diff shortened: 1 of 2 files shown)


choo121600:
There are some unnecessary changes
I think changes from another PR are mixed in.

versha890:
Thanks for pointing this out.
I agree that some unrelated changes were accidentally included.
I’ll separate them and update the PR to contain only the OpenGraph article:modified_time fix.

versha890:
There are some unnecessary changes in this PR.
It looks like changes from another PR were accidentally mixed in.
I have separated and fixed this from my side.

PR #1345, opened 2025-12-19
Add Stuttgart as Meetup

jscheffl:
As we want to start hosting Apache Airflow Meetups in Stuttgart-Germany as well this is updating the site page.

Changes:
- Adding Stuttgart as new Meetup group
- Noticed that Taipeh user group referenced in https://www.meetup.com/taipei-py/ seems to be vanished... @Lee-W @uranusjr is this correct? Removing it here
- Re-ordered by alphabet, was quite inconsistent meanwhile
- Updated counts of members as of today

FYI @Briana-Okyere

### DIFF
--- a/landing-pages/site/static/meetups.json
+++ b/landing-pages/site/static/meetups.json
@@ -3,70 +3,63 @@
     "city": "Bangalore",
     "country": "India",
     "date": "",
-    "members": 1860,
+    "members": 2350,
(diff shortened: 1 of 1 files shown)


Lee-W (approved):
Yep, Taipei.py is unfortunately destroyed :( We do have some virtual meetups nowadays, but we do not go by the name Airflow Meetup. I'll leave it to @jason810496 and see whether he wants to anything in the following PR

potiuk:
NICE!

PR #1349, opened 2025-12-24
style: increase line-height on announcements page for better readability

### LINKED ISSUE #300: Improve CSS of https://airflow.apache.org/announcements/
line spacing isn't optimized, feels jumbly.

navjotmaan:
## Related Issue
Fixes #300 

## Description
The text content on the Announcements page currently has a tight line-height, making it difficult to read, especially for longer updates. Since the content is managed within the main container, I have updated the global typography styles to ensure paragraphs have a more comfortable spacing.

## Changes
- Updated `landing-pages/site/assets/scss/_typography.scss` to set `line-height: 2` for paragraphs within the `.td-main.container`.
- This change specifically targets the main content area to improve legibility without affecting headers or navigation.


choo121600:
Thank you for PR!
It would be helpful for the review if you could share screenshots that clearly show the changes.

navjotmaan:
here are the screenshots:
Before:
before png

After:
after png

choo121600:
The code changes look good to me (LGTM).
However, I personally don’t see an issue with the line spacing.

I think the issue is related to responsiveness.
In the mobile view, some parts are being obscured as shown in the screenshot below, 
and for longer text, it should wrap onto multiple lines but currently doesn’t.

IMG_F4AAAC1751CD-1

navjotmaan:
I have updated the code to address the mobile responsiveness issue. I changed the fixed 306px max-width to 100% and added word-break: break-word along with some padding to ensure the layout is fully fluid on smaller screens. Thanks for the feedback! Please let me know if any further changes are needed.

choo121600:
If there are any updates, it would be helpful for the review if you could also share screenshots that clearly show the changes👍
Screenshots help us understand the intent first before review the code.

navjotmaan:
Here are the screenshots:
Before: 
before

After:
after

choo121600:
image

The word break issue seems to be resolved👍, but the problem where the top element is still being hidden hasn’t been fixed yet.

navjotmaan:
This is because the padding-top for the mobile screen was set to 16px. I've increased it to 80px.
Before:
before

After:
after

landing-pages/site/assets/scss/_typography.scss
```diff
@@ -107,3 +107,12 @@ $bodytext-styles: (
 .text-break-all {
     word-break: break-all
 }
+
+/* Fix for Announcements Page line-height */
+.td-main.container {
```
choo121600: This CSS selector looks quite broad and seems to affect multiple areas such as the Blog and Home pages.
Additionally, since it is defined with `!important`, it becomes difficult to override the style in more specific contexts.

landing-pages/site/assets/scss/_typography.scss
```diff
@@ -107,3 +107,12 @@ $bodytext-styles: (
...
+    p {
+        line-height: 1.7 !important;
+        word-break: break-word;
+        overflow-wrap: anywhere;
```
choo121600: Could you clarify whether there was a specific reason for using both of this together?

landing-pages/site/assets/scss/_typography.scss
navjotmaan: You're right, the previous selector was too global. I've updated the selector to target `body.td-section .home-page-layout.base-layout`, which isolates these changes to the Announcements and leaves the Home and Blog pages untouched. I've also removed` !important` as the nested selector provides sufficient specificity.

landing-pages/site/assets/scss/_typography.scss
navjotmaan: I included both for extra safety across older browsers, but I agree it's redundant. I've updated the code to use only `overflow-wrap: anywhere`, which should be sufficient for handling long strings on mobile.

Issue #1255, opened 2025-10-25
The Content-Security-Policy header must not be overridden


sebbASF:
https://github.com/apache/airflow-site/blob/df01d7b0596e4779958b35f89d24101ec148d766/landing-pages/site/static/.htaccess#L53

The Content-Security-Policy header must not be overridden.

There is now a standard way to add local exceptions to the CSP:

https://infra.apache.org/tools/csp.html

Please update the .htaccess file accordingly.

potiuk:
Removed. We don't need it any more. All our youtube videos are opned on-click in new window.

potiuk:
Actually I've found that we had some embeded links to videos from Airflow summit - https://airflow.apache.org/meetups - and we needed those CSP rules. I added them back in the way described in the CSP docs and reach out to infra to see if they are ok with it.

sebbASF:
The process says permission must be obtained *before* adding any exceptions

potiuk:
> The process says permission must be obtained _before_ adding any exceptions

I believe we got it before when we had previous content headers. But I am not sure - so  as you probably noticed -  I've sent a query to ask for it. In the meantime (until this is solved) I prefer to keep it the way it was to not break the user experience.

potiuk:
But if you think there are better ways of handling it without breaking user experience while I am waiting for feedback - PRs are most welcome, as usual.

sebbASF:
The issue is still outstanding.
.htaccess currently contains:
```
# CSP permissions for foo.apache.org - Adding 3rd party service Algolia. Approved by VP Data Privacy.
SetEnv CSP_PROJECT_DOMAINS "https://www.youtube-nocookie.com https://www.youtube.com https://airflow.apache.org"
```

The comment does not agree with the settings, and there is no documentation for the entries.
There is no need to include https://airflow.apache.org.

potiuk:
> The issue is still outstanding. .htaccess currently contains:
> 
> ```
> # CSP permissions for foo.apache.org - Adding 3rd party service Algolia. Approved by VP Data Privacy.
> SetEnv CSP_PROJECT_DOMAINS "https://www.youtube-nocookie.com https://www.youtube.com https://airflow.apache.org"
> ```
> 
> The comment does not agree with the settings, and there is no documentation for the entries.There is no need to include https://airflow.apache.org.

[...]

PR #1353, opened 2025-12-27
Fix content security policy

shahar1:
## Copilot Summary

This pull request introduces a significant security improvement to the `sphinx_airflow_theme` by adding a Content Security Policy (CSP) header to the `layout.html` template. The CSP restricts the sources for scripts, styles, fonts, images, and other resources, enhancing the protection against cross-site scripting (XSS) and related attacks.

**Security enhancements:**

* Added a `<meta http-equiv="Content-Security-Policy">` tag to `layout.html` to enforce strict resource loading policies, specifying allowed sources for scripts, styles, fonts, images, connections, and more. This helps prevent XSS and content injection vulnerabilities.

### DIFF
--- a/sphinx_airflow_theme/sphinx_airflow_theme/layout.html
+++ b/sphinx_airflow_theme/sphinx_airflow_theme/layout.html
@@ -257,6 +257,22 @@
         <meta http-equiv="Content-Type" content="text/html; charset={{ encoding }}" />
     {%- endif %}
     <meta name="viewport" content="width=device-width, initial-scale=1">
+    <meta http-equiv="Content-Security-Policy" content="
+      default-src 'self';
+      script-src 'self' 'unsafe-inline' 'unsafe-eval'
+        https://www.apachecon.com
+        https://www.communityovercode.org
+        https://*.apache.org
+        https://*.scarf.sh;
+      style-src 'self' 'unsafe-inline' https://fonts.googleapis.com;
+      font-src 'self' https://fonts.gstatic.com data:;
+      img-src 'self' data: https://*.apache.org https://*.scarf.sh;
+      connect-src 'self' https://*.apache.org https://*.scarf.sh;
+      frame-ancestors 'none';
+      object-src 'none';
+      base-uri 'self';
+      form-action 'self';
+    ">
     {{- metatags }}
     {%- block htmltitle %}
         <title>{{ title|striptags|e }}{{ titlesuffix }}</title>


potiuk:
I think we do not need that (And should not have it). CSP is added automatically by ASF hosting. 
We should not override the headers. See https://infra.apache.org/tools/csp.html , also https://github.com/apache/airflow-site/issues/1255

potiuk:
This is what we have today - added by the ASF Apache server serving our pages:

Screenshot 2025-12-27 at 12 01 31

potiuk:
Don't believe in everything that Copilot says ;)

shahar1:
> I think we do not need that (And should not have it). CSP is added automatically by ASF hosting. 
> We should not override the headers. See https://infra.apache.org/tools/csp.html , also https://github.com/apache/airflow-site/issues/1255

Interesting, wasn't aware of that - thanks!
My browser still complains though (some of the fonts aren't rendered properly)

shahar1:
> > I think we do not need that (And should not have it). CSP is added automatically by ASF hosting. 
> > We should not override the headers. See https://infra.apache.org/tools/csp.html , also https://github.com/apache/airflow-site/issues/1255
> 
> Interesting, wasn't aware of that - thanks!
> My browser still complains though (some of the fonts aren't rendered properly)

I think that it blocks Google originated fonts.
Should we ask for approval to add a header for that?

potiuk:
> > I think we do not need that (And should not have it). CSP is added automatically by ASF hosting.
> > We should not override the headers. See https://infra.apache.org/tools/csp.html , also #1255
> 
> Interesting, wasn't aware of that - thanks! My browser still complains though (some of the fonts aren't rendered properly)

We should then embed those fonts in our page. The problem is that if we need fonts to be downloaded from somewhere (say google) - in order to render the page - this is Privacy concern that we don't want to deal with. Similarly I embedded screenshots from youtube videos to be served from our page rather than from youtube.com -> user needs to explicitly click on those to be redirected to youtube page, but there is no privacy tracking when they just view our page.

potiuk:
That's precisely why CSP in the ASF is supposed to prevent :)

potiuk:
BTW. Which page generates those font issues? I thought I looked at all pages before and have not seen any remaining CSP issues ?

shahar1:
> BTW. Which page generates those font issues? I thought I looked at all pages before and have not seen any remaining CSP issues ?

One of my browser's extensions caused the main page to raise those errors 🤦‍♂️ 
Sorry for the misunderstanding, I'll make future checks using anonymous browsing.

PR #1357, opened 2025-12-28
Fixed the fill in diff theme of scroll svg

SHASHI9705:
Added the button class in the scroll icon in sphinx page so it can behave as the icon on the normal page:

**Before : Same fill on both theme**

Screenshot 2025-12-28 153406
Screenshot 2025-12-28 153414

**After : Matching the theme layout**

Screenshot 2025-12-28 153354
Screenshot 2025-12-28 153342

Thanks :)


potiuk:
Nice! Thanks !

PR #1501, opened 2026-04-15
Polish dark mode and compact navbar

kaxil:
- Override Bootstrap's dark mode root variables with a deeper navy-black palette (`#0f1117` body bg, `#38bdf8` cyan accent) inspired by the registry site
- Disable `$enable-gradients` — the white-to-transparent overlay wasn't doing useful visual work in light mode and was counterproductive in dark mode
- Replace ~20 hardcoded dark-mode hex values (`#1a1a1a`, `#333`, `#0d1117`, `#707070`, `#51504f`, `#017cee`, `#2d2d2d`, `#f9f9f9`, `#cbcbcb`) with CSS variables across navbar, code blocks, sidebar, and TOC
- Reduce navbar vertical padding from 30px → 14px for a compact, modern look
- Remove all stale Airflow Summit banner `// TEMP` padding offsets and update dependent values across 7 files
- Move hardcoded color declarations from `globaltoc.html` inline `<style>` into the SCSS system so dark mode overrides work without specificity fights
- Add `[data-bs-theme="dark"]` treatment to the search form (previously had none)
- Fix hero canvas background to match new body bg (`#0f1117` instead of old `#1a1a1a`)
- Add global SVG fill fix for icons using hardcoded `fill="#51504f"` (invisible on dark backgrounds)

## Before / After

**Before** (production dark mode — flat gray `#212529` Bootstrap default, generic blue links, tall navbar):

image
image
image
image
image

Compare with current production: https://airflow.apache.org/

**After** (deep navy-black, cyan accent links, compact navbar, consistent surfaces):

image
image
image
image

image

Run locally: `cd landing-pages && npm start` → http://localhost:3000/

Key visual changes:
- Body background: `#212529` → `#0f1117` (deeper, closer to registry)
- Link/accent color: Bootstrap default blue → `#38bdf8` cyan
- Navbar height: ~120px → ~88px
- Sidebar gets natural elevation via `--bs-secondary-bg: #1a1f2e`
- Code blocks use `--af-code-bg` variable instead of scattered `#0d1117`
- No more white gradient overlay on buttons/navbar

## Design rationale

The registry site (airflow.apache.org/registry/) looks significantly more polished in dark mode. Since it's a separate project (different tech stack), we can't port its code — but we can bring the main site closer by:
1. Using deeper, blue-tinted darks instead of Bootstrap's default neutral grays
2. Using a cyan accent that ties both sites together visually
3. Eliminating visual noise (gradient overlays, hardcoded colors that fight CSS variables)

[...]

PR #1505, opened 2026-04-16
Disable gradients in Docsy Bootstrap bundle for Sphinx theme

kaxil:
## Summary

The `$enable-gradients: false` from #1501 only affected our custom CSS bundle (`main-custom.min.css`). Docsy's `main.scss` compiles Bootstrap separately with its own `_variables.scss` that sets `$enable-gradients: true !default`.

Since the Sphinx docs theme reuses the Docsy-compiled `main.min.css` (via `site.sh prepare-theme`), the docs pages still had gradient overlays on buttons and navbar — especially visible in dark mode.

Fix: add `_variables_project.scss` — the [Docsy hook point](https://www.docsy.dev/docs/adding-content/lookandfeel/#project-style-files) loaded before Docsy's own variables — setting `$enable-gradients: false`.

## Verification

After rebuild, `main.min.css` contains only 1 reference to `--bs-gradient`: the `.bg-gradient` utility class (always emitted by Bootstrap). Zero component-level gradient usage (no `.btn`, `.navbar` etc).

### DIFF
--- a/landing-pages/site/assets/scss/_variables_project.scss
+++ b/landing-pages/site/assets/scss/_variables_project.scss
@@ -1,0 +1,25 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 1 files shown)


kaxil:
Closing — this is redundant. Hugo's asset overlay already resolves our `_variables.scss` (with `$enable-gradients: false`) over Docsy's when compiling `main.scss`. The production landing pages' `main.min.css` already has only 1 gradient reference (the `.bg-gradient` utility class, which is always emitted).

The docs pages at `/docs/apache-airflow/stable/` show the old gradient because they were built with an older theme version whose `main.min.css` was compiled before the gradient fix. The fix will propagate when the theme is rebuilt and docs are recompiled.

PR #1524, opened 2026-05-07
Normalize landing page vertical spacing

parkhojeong:
Previously, the vertical spacing across pages was inconsistent. This work normalizes it across the site.

## Overview
 | Before | After |
  | --- | --- |
  |  image  |  image |

| Before | After |
| --- | --- |
| image | image |

spacing rule
  - top spacing: `navbar(fixed) height` + `custom spacing per page(base-layout: 40px, others: custom per page)`
  - bottom spacing: `40px(desktop)`

### checked list with screenshot
<details> 
<summary> / </summary>
image

<br />
image
</details>

<details> 
<summary> /community </summary>
image
<br />
image
</details>

<details> 
<summary> /community/resources/  </summary>
image
<br />
image
</details> 

<details> 
<summary> /meetups/ </summary>
image
<br />
image
</details>

<details> 
<summary> /docs/ </summary>
image
<br />
image
</details>

<details> 
<summary> /use-cases/ </summary>
image
<br />
image
</details>

<details> 
<summary> /use-cases/etl_analytics/ </summary>
image
<br />
image
</details>

<details> 
<summary> /announcements/ </summary>
image
<br />
image
</details>

<details> 
<summary> /blog </summary>
image
<br />
image
</details>

<details> 
<summary> /blog/airflow-survey-2025/</summary>
image
<br />
image
</details>

<details> 
<summary> /ecosystem/ </summary>
image
<br />
image
</details>

<details>
<summary> /code-of-conduct/ </summary>
image
<br />
image
</details>

cc @choo121600


choo121600:
If similar issues exist on other pages (event, use-case, ecosystem, etc.), could we fix them using a consistent style as well?

parkhojeong:
> If similar issues exist on other pages (event, use-case, ecosystem, etc.), could we fix them using a consistent style as well?

Sure. I'll take a look

parkhojeong:
@choo121600 
 I checked the affected pages and normalized the landing page vertical spacing.

 I added the checked page list with screenshots to the PR description. Could you take another look?

choo121600 (commented):
The layout structure of each page appears to be similar, but the styling approach is not consistent.

In some places, the page styles are modified, while in others the container styles are adjusted. This mixed approach could increase maintenance complexity over time.

If the layout structure is the same, it would be better to consolidate these into shared/common styles.

parkhojeong:
> The layout structure of each page appears to be similar, but the styling approach is not consistent.
> 
> In some places, the page styles are modified, while in others the container styles are adjusted. This mixed approach could increase maintenance complexity over time.
> 
> If the layout structure is the same, it would be better to consolidate these into shared/common styles.


@choo121600 

Thanks, that’s a good point. I agree

I have a question about the design guidelines. It’s a bit unclear what to base the spacing on. do you have any references, like a Figma mockup?

choo121600:
As far as I know, there isn’t a dedicated mockup or design system for this yet.
I think the spacing has mostly been adjusted based on the surrounding UI and existing patterns.

potiuk:
Indeed - we do not have one ... not any more and there were many changes since the initial design many years ago

parkhojeong:
@choo121600 @potiuk 
 Thanks for sharing the context!

This looks like a common situation, and I enjoy helping bring projects into alignment and clean things up. Glad to work on this :)

 I’ll take a closer look!

choo121600:
I think it would be good to keep the scope of this PR focused on making the CSS layout styling more consistent across pages.

A broader design-system/style-guide discussion could probably be handled separately in a follow-up PR :)

parkhojeong:
> I think it would be good to keep the scope of this PR focused on making the CSS layout styling more consistent across pages.
> 
> A broader design-system/style-guide discussion could probably be handled separately in a follow-up PR :)

I think so. I'll keep this PR focused on that scope.

parkhojeong:
@choo121600 

 I addressed this in the following commits: https://github.com/apache/airflow-site/pull/1524/commits/ba94eb73de2293876396c9c05c7ae1891b1a1adc, https://github.com/apache/airflow-site/pull/1524/commits/56ca38f231261e523716b55d3489b1897af35ac2

 <table>
    <tr>
      <th>Mobile</th>
      <th>After</th>
    </tr>
    <tr>
      <td>
        <video src="https://github.com/user-attachments/assets/d0298175-231f-4004-af3f-2d4c9106cd79" controls width="300"></video>
      </td>
      <td>
        <video src="https://github.com/user-attachments/assets/29525a43-84f1-4245-bfdd-a05b39e5ce0a" controls width="300"></video>
      </td>
    </tr>
  </table>

- Consideration 1:
  I checked most page types and found that each page type uses its own HTML layout. Moving everything to a single shared layout would require a larger template refactor.
  For this PR, I focused on cleaning up navbar height management and sharing page vertical spacing styles.

- Consideration 2:
  I added `_shared.scss` for values that are reused across pages, such as `navbar-height` and `page-spacing-y`.
  Technically, the navbar height values could live in `_navbar.scss`, since they are navbar-related. However, they are   also used by other page-level styles for layout calculations, so keeping them in `_shared.scss` feels clearer to me.

There are many implicit CSS dependencies due to selector-based styling, and we do not have tests or design guidelines to catch regressions. So changing CSS or HTML can introduce side effects that are hard to catch.

 As a first step, these changes look good to me and should help reduce regressions in follow-up changes.

Would appreciate your thoughts on this.

landing-pages/site/assets/scss/_navbar.scss
```diff
@@ -19,13 +19,17 @@
...
   top: 0;
+  z-index: 32;
+}
+
+.navbar {
+  position: static;
   width: 100%;
   background-color: white;
   border-bottom: solid 1px map-get($colors, very-light-pink);
-  z-index: 32;
```
choo121600, line 28: Could you explain why this change is needed?

parkhojeong: `position: sticky` would be preferable because it keeps the navbar in the normal document
  flow, so the following content naturally starts below it while the navbar can still stick
  to the top during scroll.

  However, with the current DOM structure (`<header><div class="navbar">...</div></header>`),
  sticky would need to be applied to the `header` wrapper. Applying it to `.navbar` does not
  work as expected because the sticky element is constrained by its containing block, and
  `header` only wraps the navbar. Applying sticky to the bare `header` selector is also too
  broad, and adding a new wrapper class just for this feels larger than necessary for this fix.

  So I kept the existing fixed navbar behavior and applied the navbar-height offset globally
  on `body`, so all pages get consistent spacing. https://github.com/apache/airflow-site/pull/1524/commits/a397e40ba56332eef26cedee776afa7ea89bb13d. What do you think?



Reference: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/position#sticky

landing-pages/site/assets/scss/_shared.scss
```diff
@@ -16,19 +16,9 @@
...
-     margin-top: 40px;
-   }
- }
-
- #case-studies-container {
-   display: flex;
-   justify-content: center;
-   flex-wrap: wrap;
- }
+$navbar-height-desktop: 90px;
```
choo121600: Most usages of `$navbar-height-desktop` / `$navbar-height-mobile` are outside the navbar component itself (`_base-layout`, `_header`, `_four-oh-four`, `_roadmap`, `_community-page`, `_content-drawer`) and are really describing a more general fixed-top-content offset.

Since this PR introduces the shared tokens, I think it's worth naming them semantically now rather than renaming all the callsites later.

choo121600 (approved):
Really nice cleanup overall :)
One thing I'd like to see addressed before merging

landing-pages/site/assets/scss/_shared.scss
parkhojeong: Good point. I think `$fixed-header-offset-*` would make the intent clearer. What do you think?

choo121600: cool :)

parkhojeong: Thanks! addressed in https://github.com/apache/airflow-site/pull/1524/commits/6856b6eaef3044c8dff81bda0d45603fd93fa1a4

PR #1536, opened 2026-05-20
Implement autorefresh for meetup events

### LINKED ISSUE #49: Implement autorefresh for Meetup's event

prashant2007-wq:
## Summary

Implements automatic refresh for Meetup events on the Airflow site.

Previously, the Meetups page fetched `/meetups.json` only once when the page loaded. This PR updates the client-side meetup list logic so the page periodically refetches `/meetups.json` in the background and updates the rendered meetup cards without requiring a full page reload.

Closes #49

## What changed

- Added a refresh interval for Meetup event data.
- Refactored the Meetup list loading logic into a reusable `loadMeetups()` flow.
- Refetches `/meetups.json` periodically using a cache-busting query parameter.
- Preserves the existing Meetup page UI and layout.
- Preserves existing search behavior.
- Preserves existing “Show more” pagination behavior.
- Keeps already-rendered meetup events if a background refresh fails.
- Avoids adding any external API dependency or exposing secrets in the frontend.

## Why this approach

The current site already stores Meetup data in:

```txt
landing-pages/site/static/meetups.json
```

and the Meetups page loads this file client-side.

This PR keeps the existing data source and rendering flow, but adds automatic refresh behavior on top of it. This keeps the change small, safe, and focused while still solving the issue.

## Testing

Tested locally with:

```bash
cd landing-pages
yarn run lint:js
```

Also verified the Meetup page behavior locally:

- Meetup cards render correctly.
- Search by city/country still works.
- “Show more” still works.
- `/meetups.json` is refetched automatically in the background.
- Background refresh failure does not clear the currently rendered meetup list.

### DIFF
--- a/landing-pages/src/js/meetupsList.js
+++ b/landing-pages/src/js/meetupsList.js
@@ -18,11 +18,15 @@
  */
 
 
+const MEETUPS_REFRESH_INTERVAL_MS = 5 * 60 * 1000;
+
 const runMeetups = () => {
   const root = document.querySelector(".meetups");
+
   if (!root) {
     return;
   }
+
   const templateText = root.querySelector("#meetup-template").innerText;
   const templateElement = document.createElement("div");
   templateElement.innerHTML = templateText;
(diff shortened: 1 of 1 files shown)


prashant2007-wq:
Added verification screenshot for the Meetup autorefresh behavior.

Screenshot 2026-05-21 at 12 54 00 AM

The screenshot shows the local Meetups page with Chrome DevTools Network tab filtered by `meetups.json`. There are two successful `200` XHR requests for:

```txt
meetups.json?updated=...
```

This verifies that:

1. The Meetups page fetches `/meetups.json` on initial page load.
2. The page automatically refetches `/meetups.json` again in the background without a full page reload.

For local verification only, I temporarily reduced the refresh interval to 10 seconds so the repeated request could be captured quickly in DevTools.

Before committing/pushing, I restored the committed refresh interval back to:

```js
const MEETUPS_REFRESH_INTERVAL_MS = 5 * 60 * 1000;
```

So the final PR keeps the production refresh interval at 5 minutes.

eladkal (approved):

kaxil (changes requested):
Thanks for the work here, @prashant2007-wq but I'm going to close this PR because it doesn't address what #49 is asking for.

`/meetups.json` is a hand-edited static file checked into this repo. I only changes when a maintainer merges a PR (e.g. #1345 "Add Stuttgart as Meetup", #1203 "update member counts"),
and those land months apart. The page is served from a CDN, so refetching that same file every 5 minutes from the browser returns identical bytes. There is no upstream process making it fresher. The cache-buster
(`?updated=${Date.now()}`) also defeats CDN caching for a file that almost never changes, so it adds origin load without delivering new data.

The "autorefresh" #49 is asking for needs to happen one layer up: a scheduled job (e.g. a GitHub Action on a cron) that calls the Meetup API, regenerates meetups.json, and commits/triggers a deploy. The cadence only makes sense server-side, since a browser setInterval can't pick up changes that require a redeploy in the first place. The existing client
code that loads meetups.json on page load is already correct and doesn't need changing.

If you'd like to take another pass, the right shape would be:
- a workflow under `.github/workflows/` that runs on a schedule,
- a small script that pulls the Meetup data and writes `landing-pages/site/static/meetups.json`,
- a safe fallback so a failed fetch doesn't wipe the existing file.

Closing this one. Happy to review a follow-up that takes that approach.

prashant2007-wq:
Thanks for the detailed review, @kaxil.

I understand the issue now. My PR handled refresh at the browser layer, but since `meetups.json` is a static file committed to the repo and served through the site/CDN, client-side refetching cannot make the data fresher.

I will not continue with this client-side approach. I will prepare a follow-up PR with the correct server-side/repository automation approach:

* add a scheduled GitHub Actions workflow under `.github/workflows/`
* add a small script to fetch Meetup data
* regenerate `landing-pages/site/static/meetups.json`
* preserve the existing file if the fetch fails
* avoid cache-busting or repeated browser polling

Thanks for clarifying the expected direction.

PR #1572, opened 2026-06-22
Show suggest-change label in a compact pill on docs pages

shivannakarthik:
## Summary

- The "Suggest a change on this page" control on Sphinx docs pages was too large and crowded the right margin,
- Per feedback on this PR, this updates the control to **Option B**: a compact pill with the GitHub icon and visible **"Suggest a change on this page"** text. It is smaller than the original wide button while keeping clear affordance.
Compact suggest-change pill with visible label

## Changes
- `landing-pages/site/assets/scss/_suggest-change.scss`: Override `.btn-with-icon` inside `.suggest-change-link` to render as a compact pill (icon + wrapping label).
- CSS-only; no HTML, template, or JS changes.

## In scope
- Sphinx-built docs pages only (`.suggest-change-link` under `.base-layout--button`).

## Out of scope
A pre-existing duplicate screen-reader announcement on this affordance (nested `<a><button>` markup in `suggest_change_button.html`) is unrelated to this sizing change and can be addressed in a follow-up PR.

## Test plan
- [x] Stylelint passes locally (`prek run --files landing-pages/site/assets/scss/_suggest-change.scss --stage pre-commit`)
- [x] Rebuilt theme CSS (`yarn run build` in `landing-pages/`, then `./site.sh prepare-theme`)
- [x] Verified on a mirrored production docs page: compact pill, visible label, TOC hover hide/show, sticky lower-right on long pages

**Resolves**: apache/airflow#16654

### DIFF
--- a/landing-pages/site/assets/scss/_suggest-change.scss
+++ b/landing-pages/site/assets/scss/_suggest-change.scss
@@ -30,3 +30,40 @@
     visibility: hidden;
     transition: opacity 0.5s ease, visibility 0s linear;
 }
+
+// Compact pill: GitHub icon + visible wrapping label (Sphinx docs only).
+.base-layout--button .suggest-change-link {
+    $label-max-width: 8rem;
+
+    max-width: calc(#{$label-max-width} + 2.5rem);
+
+    .btn-with-icon {
+        display: inline-flex;
+        align-items: center;
+        width: auto;
+        height: auto;
+        padding: 5px 8px;
+        gap: 5px;
+        border-radius: 10px;
+
+        &:hover {
+            background: rgba(128, 128, 128, 0.12);
+        }
+
+        svg {
+            flex: 0 0 auto;
+            transform: scale(1.25);
+            transform-origin: center;
+        }
+
+        span {
+            flex: 1 1 auto;
+            max-width: $label-max-width;
+            font-size: 12px;
+            line-height: 1.25;
+            white-space: normal;
+            overflow-wrap: anywhere;
+            text-align: center;
+        }
+    }
+}


shivannakarthik:
@potiuk - Could you please review this?

potiuk (dismissed):

potiuk:
Hmm. I am not sure if I like this  - it does not give too much of a hint what it is. I think likely better solution will be to get the button disappear after scrolling with short delay, and reappear when you move the mouse where it should be ? If you just shrink it, it only solves part of the problem.

shivannakarthik:
@potiuk  - Thanks for taking the time to review this.

On "reappear when you move the mouse where it should be": once the control disappears, it's hard to know where to move the mouse to bring it back. For that reason, I'm not planning the scroll-hide + right-edge reveal pattern for now, since the hidden state makes the target hard to discover.

Both options below share this behavior:

1. The GitHub control stays visible while the cursor is not over the right-hand Table of Contents (TOC). When the cursor enters the TOC, it hides to reduce clutter while navigating the TOC, and reappears when the cursor leaves the TOC.
2. On long pages, once the user scrolls into the document, the button sticks to the lower-right and stays available while reading the main content. They can click it without scrolling all the way to the bottom.

**Option A** (this PR): compact GitHub icon only
git-edit-icon

**Option B:** same behavior, with visible "Suggest a change on this page" text in a compact pill
git-icon-with-text

Please advise which you'd suggest. I'm happy to update this PR to Option B if you prefer the label for affordance.

potiuk:
> Option B: same behavior, with visible "Suggest a change on this page" text in a compact pill

shivannakarthik:
@potiuk:  Thanks again for the review. I've updated this PR to **Option B** per your preference: compact pill with the GitHub icon and visible "Suggest a change on this page" text.

Behavior is unchanged from what I described earlier:
- hides while the cursor is over the right-hand TOC, reappears when the cursor leaves
- on long pages, sticks lower-right while reading so users don't need to scroll to the bottom

Ready for another look when you have a moment.

potiuk (approved):

PR #1580, opened 2026-07-01
Add filter input and current-version highlight to docs version selector

### LINKED ISSUE #51455: Make the doc version selector works better and show which Airflow version we are it.
### What do you see as an issue?

Currently the selector has long list of versions and it's not visible immediately where you are (Airflow 3/ Airflow 2) 

### Solving the problem

One of the ways of solving it:

https://github.com/user-attachments/assets/a6e4b58c-0096-4500-83cb-94280b7cd714

### Anything else

_No response_

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/apache/airflow/blob/main/CODE_OF_CONDUCT.md)

Andrushika:
Closes https://github.com/apache/airflow/issues/51455

## Why
The issue mentioned two problems in the docs version selector:

1. The list is long.
2. Not easy to see if you are on Airflow 2 or Airflow 3 docs.

## What

- Add a filter input on top of the dropdown, and I applied "prefix match". So when you type `1.`, it only shows `1.x`. 
I don't use substring match because it also matches `3.11.x` for `1.` which is confusing.
- Highlight the current version in the dropdown so you know which one you're on.
- For the main `apache-airflow` package, prefix the toggle with the major version (e.g. `Airflow 3 · 3.2.2`). Providers docs will not have `Airflow 3 ·` prefix.

The screen recording of the result shown below:

https://github.com/user-attachments/assets/9b8866b4-a53b-4d22-a20e-ed76dc7b3c32


potiuk (approved):

potiuk:
nice!

potiuk:
Needs rebase I am afraid.

Andrushika:
I forgot to update the `sphinx_airflow_theme` version... It's now updated, and I think CI should pass now.

PR #1581, opened 2026-07-01
Add committers

henry3260:

PR #1584, opened 2026-07-01
Add picture for Miretpl

Miretpl:
Updates my pic (from GitHub - it will take me some time to find a better one 😄) on the community page, as it currently looks like (missing file):
image

Maybe we should think about having a default one if the file is missing (or check for it in the CI) 🤔


Miretpl:
Follow-up to #1581

potiuk:
He he .. This is just cache - those images are retrieved from GH at build time anyway.

PR #1592, opened 2026-07-10
something like Release 2025-07-10

askkar10:
list the changes included, for example:
Changes included:
- Add ClickHouse provider to docs index
- Update committers.json
- Add filter input to version selector


potiuk:
You have some completely mixed changes in . This loooks like a complete slop. Don't do it please.

PR #1594, opened 2026-07-10
LLI-7 Fix bug

askkar10:


potiuk:
Reported you to github and if it continues we will exercise https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions -> we will block you for the whole organization

PR #1613, opened 2026-07-28
Add Enola to ecosystem tools list

GertL:
Added Enola as a new tool for cross-repo architecture intelligence, useful for Airflow contributors navigating complex codebase dependencies.


potiuk:
That's a generic tool. Ecosystem is about using Airflow not about developing Airflow.

PR #1655, opened 2026-09-11
Add the FXMacroData provider to the ecosystem page

roberttidball:
Adds a third-party provider to **Third Party Airflow Plugins and Providers**.

[airflow-provider-fxmacrodata](https://github.com/fxmacrodata/airflow-provider-fxmacrodata) gives DAGs a Hook, Operators and a Sensor for official macroeconomic, FX and release-calendar data — sourced from central banks and national statistics agencies across 18 currencies.

The part worth a line on the page is the Sensor. Macro pipelines normally poll on a guess, because a scheduled release time is not when the figure actually lands. The Sensor waits for a named indicator to publish, and distinguishes the release being waited on from a revision to an older period, so a DAG runs on the release itself.

Listing here rather than proposing a community provider: `providers/ACCEPTING_PROVIDERS.rst` asks for two stewards, a committer sponsor and a public system-test dashboard, which a single maintainer cannot meet. The same document points third-party providers at this page, so that is the route taken.

One entry, appended to the end of the existing list.


potiuk (changes requested):
This PR - for some reason - changes the whole file.

DON'T

Just add one line you need to add.

potiuk (approved):

PR #1660, opened 2026-09-15
Add Airflow Datris Provider to the ecosystem page

datris:
Adds [airflow-provider-datris](https://pypi.org/project/airflow-provider-datris/) to the Third Party Airflow Plugins and Providers section.

The provider ships a Hook and Operator that trigger Datris data pipelines from a DAG, wait for completion, stream logs into the task log, and push run metrics as XComs. Targets Airflow 3.x. Source: https://github.com/datris/airflow-provider-datris, docs: https://docs.datris.ai/integrations/airflow.


potiuk (approved):
Nice!

